PR #6, opened 2016-11-23
MIR-527 added empty </cc:address>

kkrebs:


kkrebs:
please leave this branch for more complex solution, this is only a first version

PR #17, opened 2016-11-30
MIR-535 break file name in file browser if to long

paschty:


kkrebs:
please check if there are other classes involved (audio and video eg.)
with class "file_name_video" this fix doesn't work atm

PR #23, opened 2016-12-09
MIR-540 Use Yarn instead of npm for mir-layout build

paschty:


yagee-de:
resolve conflicts

pom.xml
```diff
@@ -210,6 +212,13 @@
           <groupId>com.github.eirslett</groupId>
           <artifactId>frontend-maven-plugin</artifactId>
           <version>1.3</version>
+          <version>${frontend.maven.plugin.version}</version>
```
yagee-de: two `<version>` tags are not allowed

PR #25, opened 2016-12-12
MIR-545 use grunt release 1.0.0 now and also updated other grunt libs

kkrebs:


kkrebs:
rebased branch to 2016.06.0.x

PR #46, opened 2017-01-19
MIR-548 User failed to see video selection box

eckstueck:
- added panel around videoplayer and videochooser, for better visibility


eckstueck:
:ship: ?

PR #60, opened 2017-03-21
MIR-581 show state badge on metadata view

eckstueck:
added state badge to metadata view

### DIFF
--- a/mir-module/src/main/resources/xsl/metadata/mir-abstract.xsl
+++ b/mir-module/src/main/resources/xsl/metadata/mir-abstract.xsl
@@ -88,6 +88,17 @@
             <xsl:with-param name="query" select="concat('&amp;fq=link:*',$accessCondition, '&amp;owner=createdby:', $owner)" />
           </xsl:call-template>
         </xsl:if>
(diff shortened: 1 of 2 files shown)


mir-module/src/main/resources/xsl/metadata/mir-abstract.xsl
```diff
@@ -88,6 +88,17 @@
             <xsl:with-param name="query" select="concat('&amp;fq=link:*',$accessCondition, '&amp;owner=createdby:', $owner)" />
           </xsl:call-template>
         </xsl:if>
+          <xsl:if test="not(mcrxsl:isCurrentUserGuestUser())">
```
yagee-de: Isn't this information useful for all users? Why do I need to login for it?

kkrebs: because the badge on result list is also only shown for logged in users. Guests should only see published documents.

yagee-de: but isn't this stylesheet used for the metadata page and not for the result list?

PR #124, opened 2017-07-04
Issues/mir 547 use new mycore pi component for urn

kkrebs:

PR #126, opened 2017-07-06
MIR-623 remove mods:identifier from mods copy

kkrebs:


paschty:
You can check if is controlled by MyCoRe with isManagedPI of org.mycore.mir.pi.MCRPIUtil in #124

kkrebs:
we need no such check here. each publication has it's own identifier.

PR #151, opened 2017-08-09
MIR-636 use mir.genre instead of mir.type (reindex needed)

kkrebs:


mir-module/src/main/resources/xsl/response-mir.xsl
```diff
@@ -239,8 +239,8 @@
     <xsl:variable name="mcrobj" select="." />
     <xsl:variable name="mods-type">
```
yagee-de: change variable name, too

PR #179, opened 2017-10-19
MIR-671 add 'institute' OAI Set

yagee-de:
top-level categories only


yagee-de:
depends on https://github.com/MyCoRe-Org/mycore/pull/415

PR #193, opened 2017-12-04
MIR-679 fix new version listing on blocked or deleted documents

yagee-de:
mods.part was used as a sort field which is multivalued
updated yarn.lock as old version caused errors


MyCoRe-Travis:
Something gone wrong with 97dfe41275521e76dba7853ac01ab2d8d7e51275! 
 Please Check https://github.com/MyCoRe-Travis/MIR_test_artifacts/tree/1094-pull_request-193

PR #213, opened 2018-01-31
MIR-681 fix encoding for abstract/title with html content

adlerre:


yagee-de:
Bug is marked as fixed for LTS branch, but this PR only fits master and needs a rebase

PR #219, opened 2018-02-13
MIR-700 added missing & parameter

kkrebs:


mir-module/src/main/resources/META-INF/resources/js/mir/base.js
```diff
@@ -236,9 +236,9 @@
           if (origSearchAction.includes('servlets/solr/find')) {
             var replAction = origSearchAction.replace(/(.*[&|\?])(condQuery=.*?)&(.*)/,'$1$3&');
             if ($('#search_type_button').attr('value') == 'all') {
-                var newAction = replAction + "condQuery=" + addValue;
+                var newAction = replAction + "&condQuery=" + addValue;
```
yagee-de: isn't that '&' appended on line 237: '$1$3&'?
So it seems to be double appended now...

PR #227, opened 2018-02-16
MIR-710 make language repeatable for all editor forms

yagee-de:


kkrebs:
shouldn't this be repeatable in all editor masks? Why only in research data?

PR #286, opened 2018-08-06
MIR-751 updated tomcat version and encode invalid chars in index browser

kkrebs:
[Link to MIR-751](https://mycore.atlassian.net/browse/MIR-751).


kkrebs:
Can you please take a look at this? Using encoded version of `[` and `]` seems to broke my solr response:
`
<str name="terms.regex">[a|Ã¤|Ã].*</str>
`
Any idea how we can prevent this?

PR #341, opened 2018-12-12
MIR-777 Thumbnail in basket view is missing

eckstueck:
- added thumbnails in basket view
- changed thumbnails in list view


eckstueck:
Will be fixed with https://github.com/MyCoRe-Org/mycore/pull/740

kkrebs:
MCR-1997 fixed this bug, but there are still problems if object contains more than one derivates

kkrebs:
seems that this is another bug ... so I close this ticket and open a new one

PR #390, opened 2019-08-09
MIR-842 fix layout issues caused by bootsrap 4 migration

TiloNN:
Added some colours for the badges.
Each batch should have a colour now, but I did not thought about which one.

Furthermore I fixed the empty search in the navbar.
By the way I added some comments to the base.js:
I had trouble to understand the code and left some of my findings.

[Link to jira](https://mycore.atlassian.net/browse/MIR-842).

### DIFF
--- a/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_metadata.scss
+++ b/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_metadata.scss
@@ -55,25 +55,26 @@
   }
   .doc_state {
     display: inline;
-    color: $dark_neutral;
+    color: $white;
     .mir-published {
-      background-color: darken($success, 10%)!important;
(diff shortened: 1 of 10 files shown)


mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_metadata.scss
```diff
@@ -55,25 +55,26 @@
...
     .mir-review {
-      background-color: darken($primary, 10%)!important;
+      background-color: $primary;
     }
     .mir-blocked {
-      background-color: darken($danger, 10%)!important;
+      background-color: $danger;
     }
     .mir-deleted {
-      //background-color: $label-default-bg;//TODO: not sure if this variable has a equivalent in bs4
```
TiloNN: From my point of view there is none equivalent.

PR #400, opened 2019-09-17
MIR-842 copy current issue url fix layout issues caused by bootsrap 4 migration

TiloNN:
Fixes minor layout issues, caused by bootstrap 4:
- removed global margin right from buttons, caused by an outdated css class and custom rule
- fixes side effect by adding space to login buttons
- repairs the modal dialogue for document version history e.g.

### DIFF
--- a/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/forms.scss
+++ b/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/forms.scss
@@ -17,10 +17,6 @@
   top: 0;
 }
 
-*:not(.input-group-btn)/*:not(.btn-group)*/ > .btn {
-  margin-right: 5px;
-}
-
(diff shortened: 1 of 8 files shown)


fluetze:
The changes in mir-abstract.xsl won't work this way:
Once an attribute has been written by xsl it can't be changed afterwards, afaik. So, for example,

instead of 
  <li class="float-right nav-item">
                    <xsl:if test="position()=1">
                      <xsl:attribute name="class">active float-right nav-item</xsl:attribute>
                    </xsl:if>

use
<li>
                    <xsl:attribute name="class">
                      <xsl:if test="position()=1">active </xsl:if>
                      <xsl:text>float-right nav-item</xsl:text>
                    </xsl:attribute>

kkrebs:
both solutions work:

> If two attributes are output with the same name, the second one takes precedence.

PR #401, opened 2019-09-18
MIR-842 update pagination styles

adlerre:
[Link to jira](https://mycore.atlassian.net/browse/MIR-).

### DIFF
--- a/mir-module/src/main/resources/xsl/response-mir-utils.xsl
+++ b/mir-module/src/main/resources/xsl/response-mir-utils.xsl
@@ -26,7 +26,7 @@
         </xsl:variable>
         <div class="col-12 col-md-5 text-left">
           <a tabindex="0" class="previous" href="{$link}" data-pagination=".caption:mods.title.main">
-            <span class="fas fa-chevron-left icon" />
+            <span class="fa fa-chevron-left icon" />
             <span class="caption">
               <xsl:value-of select="$label.previousHit" />
             </span>
@@ -56,7 +56,7 @@
             </xsl:variable>
             <xsl:value-of select="concat($proxyBaseURL, $HttpSession, $params, '&amp;start=', $newStart, '#hit_', $page)" />
           </xsl:attribute>
-          <span class="fas fa-chevron-up" />
+          <span class="fa fa-chevron-up" />
         </a>
         <xsl:text>&#160;</xsl:text>
         <xsl:value-of select="i18n:translate(concat($i18nprefix, '.entriesInfo'), concat($page, ';', $pages))" />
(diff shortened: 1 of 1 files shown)


mir-module/src/main/resources/xsl/response-mir-utils.xsl
```diff
@@ -26,7 +26,7 @@
         </xsl:variable>
         <div class="col-12 col-md-5 text-left">
           <a tabindex="0" class="previous" href="{$link}" data-pagination=".caption:mods.title.main">
-            <span class="fas fa-chevron-left icon" />
+            <span class="fa fa-chevron-left icon" />
```
TiloNN: fa is not longer supported by current font awesome library version 5
chevron left is part of the solid font lib (fas) now
So why do you like to redo this font awesome migration?

mir-module/src/main/resources/xsl/response-mir-utils.xsl
```diff
@@ -132,7 +132,7 @@
     <xsl:variable name="label.previousPage" select="i18n:translate(concat($i18nprefix, '.previous'), $page - 1)" />
     <xsl:variable name="label.nextPage" select="i18n:translate(concat($i18nprefix, '.next'), $page + 1)" />
 
-    <ul id="{$id}-paginate" class="pagination {$class} d-inline-flex">
```
TiloNN: Seems the combination of text-center of the parent element and display inline-block of the list element is not longer possible, because pagination was set to display flex. To prevent the full width of the list and to enable centred alignment it has to be a inline element again. Someone used the d-inline-flex class for that. If you remove this, the pagination is not longer centred.

mir-module/src/main/resources/xsl/response-mir-utils.xsl
```diff
@@ -152,12 +152,12 @@
           </xsl:when>
           <xsl:otherwise>
             <xsl:attribute name="class">page-item disabled</xsl:attribute>
-            <a class="page-link"><span>
```
TiloNN: Hm, don't now why there is a span.
But I also can't imagine why you like to remove it.

mir-module/src/main/resources/xsl/response-mir-utils.xsl
```diff
@@ -345,16 +345,16 @@
...
+        <a class="page-link" tabindex="0" href="#">
           <xsl:value-of select="$paginateStart" />
-        </span></a>
+        </a>
       </li>
     </xsl:if>
     <xsl:if test="$paginateStart != number($page)">
       <li class="page-item">
         <xsl:if test="($paginateStart &lt; $paginateStart-mobile) or ($paginateStart &gt; $paginateEnd-mobile)">
-          <xsl:attribute name="class">d-xs-none</xsl:attribute>
```
TiloNN: Right, d-xs-none is not available.
Looks good.

adlerre:
Close PR, because was already fixed. But not merged to master.

PR #403, opened 2019-09-19
MIR-842 migrate XEditor templates to BS 4

adlerre:
[Link to jira](https://mycore.atlassian.net/browse/MIR-).


mir-module/src/main/resources/xsl/xeditor-mir-templates.xsl
```diff
@@ -213,7 +224,7 @@
   </xsl:template>
 
   <xsl:template match="mir:template[@name='selectInput']" mode="widget">
-    <select class="form-control input-{$input-size} {@class}" id="{@id}">
+    <select class="form-control custom-select input-{$input-size} {@class}" id="{@id}">
```
kkrebs: class="custom-select" leads to layout bug.

We will use this later, but atm it does not work because there needs more to be done.

adlerre: was removed

PR #420, opened 2019-11-08
MIR-859 security fixes

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MIR-859).


paschty:
That are just versions bumps. Do they really contain any needed security fixes?

kkrebs:
For me this is ok. But it should work without migration. I will test it next week.

yagee-de:
> For me this is ok. But it should work without migration. I will test it next week.

Will wait for the final verdict.

PR #447, opened 2020-03-11
MIR-887 add function to prevent duplicated journals in JATS2MODS

eckstueck:
added function to prevent duplicated journals in JATS2MODS

[Link to jira](https://mycore.atlassian.net/browse/MIR-887).


eckstueck:
fixed

PR #448, opened 2020-03-11
MIR-888 fixes losing classes for active property nav entry

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-888).


TiloNN:
wrong branch name

PR #449, opened 2020-03-11
MIR-889 fix classes for current user menu entry

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-889).


TiloNN:
wrong branch name

PR #457, opened 2020-04-01
MIR-893 makes dropdown bs4 conform and removes proprietary styles

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-893).

Changes are more than optimizing main menu, but it's about making dropdown menu bootstrap 4 conform. We can drop a lot of styles by using the defaults from the bootstrap and bootswatch theme.

This has probably an impact to existing applications.


yagee-de:
@eckstueck: Could you please check this changes with dbt?

PR #469, opened 2020-05-06
MIR-894 changes doctype to html and removes script type attributes

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-894).

- changes doctype to html (5)
- removes some script declarations
- fixes invalid aria attribute

### DIFF
--- a/mir-layout/src/main/resources/xsl/mir-flatmir-layout-utils.xsl
+++ b/mir-layout/src/main/resources/xsl/mir-flatmir-layout-utils.xsl
@@ -37,14 +37,14 @@
             class="navbar-toggler"
             type="button"
             data-toggle="collapse"
-            data-target="#mir-main-nav__entries"
-            aria-controls="mir-main-nav__entries"
+            data-target="#mir-main-nav-collapse-box"
+            aria-controls="mir-main-nav-collapse-box"
             aria-expanded="false"
             aria-label="Toggle navigation">
             <span class="navbar-toggler-icon"></span>
           </button>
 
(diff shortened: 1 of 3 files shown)


mir-layout/src/main/resources/xsl/mir-flatmir-layout-utils.xsl
```diff
@@ -37,14 +37,14 @@
...
-            data-target=".mir-main-nav__entries"
+            data-target="#mir-main-nav__entries"
             aria-controls="mir-main-nav__entries"
             aria-expanded="false"
             aria-label="Toggle navigation">
             <span class="navbar-toggler-icon"></span>
           </button>
 
-          <div class="collapse navbar-collapse mir-main-nav__entries">
+          <div id="mir-main-nav__entries" class="collapse navbar-collapse mir-main-nav__entries">
```
kkrebs, line 47: do we need both here? mir-main-nav__entries as ID and as class?

TiloNN: We do not need any IDs here.
But aria does and that's why I just added an ID and switched the data-target.
Short: We need the ID only for the collapse behaviour ... now, to provide valid HTML.

TiloNN: ok .. the same name for ID and class is confusing, I agree.
I changed the name of the ID. Hopefully into something more understandable.

PR #487, opened 2020-06-24
MIR-916 cosmol doesn't not work with bs4

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-916).


yagee-de:
fixed commits in #493

PR #518, opened 2020-08-24
MIR-936 add mime type and output format configuration for csl

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-936).


kkrebs:
is part of #521

PR #526, opened 2020-09-10
MIR-939 do not include migration commands by default

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MIR-939).


fluetze (changes requested):
I'd prefer to keep the last migration commands still active.
When they are included by default, I initually see the commands, run them, then disable them locally.
When they are nit included by default, I might not recognize them, must include them, run them, then exclude them again. More work, more space for errors.

yagee-de:
> I'd prefer to keep the last migration commands still active.
> When they are included by default, I initually see the commands, run them, then disable them locally.
> When they are nit included by default, I might not recognize them, must include them, run them, then exclude them again. More work, more space for errors.

How can you exclude them afterwards without loosing future updates? As you can only append to a property and not remove a part I was for exclude by default.

kkrebs (approved):

fluetze (approved):

kkrebs:
after short discussion we decided, that current migration command should be activated by default.

PR #554, opened 2020-11-30
MIR-961-search_entity_does_not_associate_identifier

rochpaul:
Adapt mir:role.repeated template and generalize queries in
search-entity.js

[Link to jira](https://mycore.atlassian.net/browse/MIR-961).

### DIFF
--- a/mir-module/src/main/resources/META-INF/resources/js/jquery.search-entity.js
+++ b/mir-module/src/main/resources/META-INF/resources/js/jquery.search-entity.js
@@ -760,7 +760,7 @@
         $outputNameType.val(item.type.toLowerCase());
 
         /* Get dependent personExtended_box */
-        var itemPersonExtendedBox = $($output).closest('fieldset[class="personExtended_box"]');
+        var itemPersonExtendedBox = $($output).closest('[class="personExtended_box"]');
 
         /* get the next free output field */
         nameIdFields = $(itemPersonExtendedBox).find('input[name*="/mods:nameIdentifier"]');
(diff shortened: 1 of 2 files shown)


toKrause:
I've look at the changes and incorporated them into my application. A first search work as expected, but all subsequent searches have the same problem as before. If I'm correct, It's necessary and sufficient to remove the explicit `select`-element name in line [777](https://github.com/MyCoRe-Org/mir/blob/b555ab904015d55d32955638dbe1e9b07da295c9/mir-module/src/main/resources/META-INF/resources/js/jquery.search-entity.js#L777) as well, i.e. change it to:

```
nameIdTypesElements = $(itemPersonExtendedBox).find('[name*="/mods:nameIdentifier"]');
```
But I still haven't exactly understood whats happening in the `updateOutput`-method, so please take this with a grain of salt and test for yourself.

rochpaul:
Hello Thorsten,
thank you for your review and the advice.
I have checked the described behavior and I can confirm it. 
With another search (the error does not appear if you change the identifier type) the underlying id of selected search entity will be ignored.
There is thrown an exception on renewed selection in js console:
`
jquery.search-entity.js:821 Uncaught TypeError: Cannot read property 'name' of undefined
    at a.updateOutput (jquery.search-entity.js:821)
    at HTMLAnchorElement.<anonymous> (jquery.search-entity.js:709)
    at HTMLAnchorElement.dispatch (jquery.min.js:2)
    at HTMLAnchorElement.v.handle (jquery.min.js:2)
`

I will fix this as soon as possible.
Best regards,
Paul

rochpaul:
Hi Thorsten,
the issue should be fixed.
Please can you verify and confirm?

Best regards,
Paul

kkrebs (approved):
🚢

PR #562, opened 2021-01-20
MIR-967 use property MIR.HostedPeriodicals.List to specify list of in

kkrebs:
... repository hosted periodicals

[Link to jira](https://mycore.atlassian.net/browse/MIR-967).

Here an example for property usage:

MIR.HostedPeriodicals.List=zimport_mods_00000280,zimport_mods_00000282,zimport_mods_00000283,zimport_mods_00000284,zimport_mods_00000286,zimport_mods_00000288,zimport_mods_00000285,clausthal_mods_00000082


rsteph-de:
Is a static list in a property the right way to configure this list?
It should be possibile to detect this from journal metadata itself.

yagee-de:
A journal, series, book etc. is hosted, if it has a MCRObjectID (relatedItem is an extra object) AND it has OR has not a label in some form. Configuring a property is not a good solution as it requires a restart of the application to be considered.

PR #565, opened 2021-01-26
MIR-968 add information box for mods:person

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-968).


fluetze (dismissed):
I'd like to have this box functionality in mycore-mods, not just mir. UBO also needs that, and I already have another application with the same requirement.

kkrebs (approved):

kkrebs:
> 
> 
> I'd like to have this box functionality in mycore-mods, not just mir. UBO also needs that, and I already have another application with the same requirement.

We should clarify this, but meenwhile I like to merge this to MIR.

PR #601, opened 2021-04-22
MIR-1002 refrain from causing invalid mods:namePart elements

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1002).


kkrebs (changes requested):
This could lead to unknown side effects. It seems that we use this construct for conference names. For more details see my comment in MIR-1002.

PR #623, opened 2021-05-28
MIR-1022 MIRStrategy does not work right and is not testet

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1022).


kkrebs (approved):
tested and looks good for me

PR #627, opened 2021-06-09
MIR 1025 Option to enable mir_licences explicitly for mir_genres

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1025).

### DIFF
--- a/mir-module/src/main/resources/META-INF/resources/editor/editor-includes.xed
+++ b/mir-module/src/main/resources/META-INF/resources/editor/editor-includes.xed
@@ -1420,7 +1420,7 @@
             <option value="">
               <xed:output i18n="mir.select" />
             </option>
-            <xed:include uri="xslStyle:items2options:classification:editor:-1:children:mir_licenses" />
+            <xed:include uri="xslStyle:license-filter,items2options?genre={$genre_intern}:classification:editor:-1:children:mir_licenses" />
           </select>
         </div>
         <mir:help-pmud help-text="{i18n:mir.help.rights}" />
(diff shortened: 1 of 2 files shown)


kkrebs (changes requested):
Works as designed, but please go back to the version with only one xsl:if statement. This isn't that complex and if you add some comment as documentation to this codeblock everything is fine.

fluetze:
You could remove the <xsl:template match="@* | node()"> and simply include <xsl:include href="copynodes.xsl" />

fluetze (changes requested):
You could remove the <xsl:template match="@* | node()"> and simply include <xsl:include href="copynodes.xsl" />

fluetze (changes requested):
You could simplify the template match="items", just use
  <xsl:template match="items[string-length($genre) &gt; 0]" priority="1">
        <xsl:copy>
          <xsl:apply-templates select="@* | node()"/>
        </xsl:copy>
  </xsl:template>

golsch:
> With use of <xsl:include href="copynodes.xsl" />
> you can even simpler remove the complete xsl:template match="items" without any functional change. It is useless.

I don't know how far the interpreter optimizes the code. The `items` template is meant as an optimization, so that the checks in `item` can be saved if no genre exists.

fluetze:
> 
> 
> > With use of <xsl:include href="copynodes.xsl" />
> > you can even simpler remove the complete xsl:template match="items" without any functional change. It is useless.
> 
> I don't know how far the interpreter optimizes the code. The `items` template is meant as an optimization, so that the checks in `item` can be saved if no genre exists.

I understand, may be you can save some nanoseconds...

fluetze (approved):

kkrebs (approved):

PR #655, opened 2021-08-19
MIR-1042 exclude abstract from scopus import for license reasons

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1042).

### DIFF
--- a/mir-module/src/main/resources/xsl/import/scopus2mods.xsl
+++ b/mir-module/src/main/resources/xsl/import/scopus2mods.xsl
@@ -23,8 +23,6 @@
       </mods:originInfo>
       <xsl:apply-templates select="item/bibrecord/head/citation-info/author-keywords/author-keyword" />
       <xsl:apply-templates select="scopus:language[@xml:lang]" />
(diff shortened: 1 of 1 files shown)


fluetze:
Why did you remove
<xsl:template match="scopus:coredata/scopus:openaccess">
We need that.

In case the article is OA, the abstract is OA, too and could be preserved.

kkrebs:
atm we have no OA classification yet in MIR. We could readd this later ... In scopus license they make no difference between non oa abstracts and oa abstracts. We have to sort this out, but for now we have the demand to remove abstract scopus import.

fluetze:
Ok, I see that UBO has its own version of the file.

... but why not 
<xsl:apply-templates select="item/bibrecord/head/abstracts/abstract[current()/scopus:coredata/scopus:openaccess='1']" />
?

kkrebs:
As discussed - we should ask scopus here. On there license page there is no difference between oa abstract and non oa abstract.

sebhofmann (approved):

PR #660, opened 2021-08-26
MIR-1052 Several small issues regarding the TOC

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1052).

### DIFF
--- a/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_metadata.scss
+++ b/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_metadata.scss
@@ -259,16 +259,18 @@ $grayDark: #ECF0F1 !default;
 .mir-toc-section-toggle {
   color: #999;
   margin-right: 1ex;
+  vertical-align: top;
 }
 .mir-toc-section-toggle:hover{
   text-decoration: none;
 }
 .mir-toc-section-label {
+  display: inline-block;
   margin-right: 1ex;
 }
 .mir-toc-section-list {
   list-style: none;
-  padding-left: 2rem;
+  padding-left: 0;
   padding-bottom: 0.5rem;
 }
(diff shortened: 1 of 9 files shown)


kkrebs (approved):
looks good for me, but perhaps can fluetze also take a look at this

fluetze (commented):
Regarding solr-fields4toc[-3].xsl:

Condition as it was before (look at the braces):
<xsl:apply-templates select="(descendant::mods:relatedItem[@type='host']/mods:part/mods:detail[@type='volume'])[1]/mods:number" mode="toc.field" />
==> was intended to select the first volume info in any host, upwards.

Condition as it is now:
<xsl:variable name="allHostsParts" select="descendant::mods:relatedItem[@type='host']/mods:part"/>
==> selects all  parts in any host, upwards, ok.
<xsl:apply-templates select="$allHostsParts/mods:detail[@type='volume'][1]/mods:number" mode="toc.field" >...
==> selects the first detail of type volume, ok, 
BUT:
==> I'm not sure if this is the same as
<xsl:apply-templates select="($allHostsParts/mods:detail[@type='volume'])[1]/mods:number" mode="toc.field" >...

May be it is the same and the braces are obsolete.
Actually, we want to find
descendant::mods:relatedItem[@type='host'][mods:part/mods:detail[@type='volume']]
and from all these hosts containing a volume info , the first of them

Did you, or can you try it?

fluetze (approved):

toKrause:
the parenthesis are indeed necessary here, fixed

toKrause:
@kkrebs due to the nature of this PR, squashing is probably not helpful if not detrimental in this case. each commit deals with a separate small issue

kkrebs:
I saw this, but we have decided to have only one commit per issue finally. I put it on the agenda of our telco, we should talk about this there.

PR #675, opened 2021-09-30
MIR-1063 Refrain from adding history modal for non-privileged users

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1063).

### DIFF
--- a/mir-module/src/main/resources/xsl/metadata/mods-metadata-page.xsl
+++ b/mir-module/src/main/resources/xsl/metadata/mods-metadata-page.xsl
@@ -79,6 +79,7 @@
         <!-- this function is bloated due backward compatibility -->
         <xsl:choose>
           <xsl:when test="$boxID='mir-historydata'">
+            <xsl:if test="$originalContent/@write">
             <div
                 id="historyModal"
                 class="modal fade"
@@ -117,6 +118,7 @@
                 </div>
               </div>
             </div>
+            </xsl:if>
           </xsl:when>
           <xsl:when test="$boxID='mir-collapse-files'">
             <div class="detail_block">


mir-module/src/main/resources/xsl/metadata/mods-metadata-page.xsl
```diff
@@ -79,6 +79,7 @@
         <!-- this function is bloated due backward compatibility -->
         <xsl:choose>
           <xsl:when test="$boxID='mir-historydata'">
+            <xsl:if test="$originalContent/@write">
```
kkrebs, line 82: why should only user with write permission see this section?

kkrebs (approved):

mir-module/src/main/resources/xsl/metadata/mods-metadata-page.xsl
toKrause: long story, ...

this section, `mir-historydata`, represents a modal dialog. the only way, for a (normal) human user, to actually see this modal dialog is to press a button that is part of the `mir-admindata` section. so it makes no sense to add this section, unless the `mir-admindata` section is also added.

the proposed condition is the same condition that is used to decide whether the `mir-admindata` may be included (https://github.com/MyCoRe-Org/mir/blob/c83f1c76cf34fc70f08fb216fce9e2faae84661f/mir-module/src/main/resources/xsl/metadata/mods-metadata-page.xsl#L131)

this is not an ideal solution. an ideal solution would include `mir-historydata` if and only if `mir-admindata` is included. but as it stands, inclusion of both section is independently configurable using `MIR.Layout.[Top|Start|End|Bottom]`; presumably as an simplification. for consistency, both section should have the same additional condition, i.e. `$originalContent/@write`. 

this is a pragmatic solution that was quickly implementable.

why was a quick implementation necessary? publicly including the section publicizes the links to old revisions to non-human users, i.e web crawlers which will then request those pages. this leads to unnecessary system loads, may come withe SEO penalties (for having a lot of pages with similar content) and, worst of all, leads to a lot of error responses in combination with the issue, i raised last thursday on the mailing list (impossibility of viewing old revisions of objects with registered PIs and derivatives that have changed in the meantime).

kkrebs: I see. We should open a separate issue to keep this in mind. We need a better possibility to configure rights on these boxes.

PR #699, opened 2021-12-14
MIR-1093 enable access key filter for object servlet

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1093).


golsch:
should be merged together with https://github.com/MyCoRe-Org/mycore/pull/1399.

PR #707, opened 2022-01-20
MIR-1101 thumbnail derivate should be preferred in search results

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1101).

### DIFF
--- a/mir-module/src/main/resources/xsl/response-mir.xsl
+++ b/mir-module/src/main/resources/xsl/response-mir.xsl
@@ -328,32 +328,44 @@
     <xsl:variable name="completeHref">
       <xsl:variable name="q">
         <xsl:call-template name="detectSearchParam">
-          <xsl:with-param name="join" select="'&amp;passthrough.'" />
+          <xsl:with-param name="join" select="'&amp;passthrough.'"/>
         </xsl:call-template>
       </xsl:variable>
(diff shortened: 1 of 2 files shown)


rsteph-de:
Just for the records: On RosDok (before switching to IIIF-Thumbnail-API) I used an extra solr field "cover_url" and decided which derivate / file to display during indexing. 
It contained the complete URL-Path (without baseURL) to the image - so it was very easy to include into the search result page.

mir-module/src/main/resources/xsl/response-mir.xsl
```diff
@@ -452,136 +464,102 @@
...
+                  <div class="hit_icon"
+                       style="background-image: url('{$WebApplicationBaseURL}api/iiif/image/v2/thumbnail/{$identifier}/full/!300,300/0/default.jpg');">
+                  </div>
+                </xsl:when>
+                <!-- when there is no content derivate then use disabled icon -->
+                <xsl:when test="string-length($derivid)=0">
+                  <img class="hit_icon" src="{$WebApplicationBaseURL}images/icons/icon_common_disabled.png"/>
+                </xsl:when>
+                <xsl:otherwise>
+                  <!-- if not, then the content type decides a icon -->
```
kkrebs, line 528: please readd missing code line
<div class="hit_icon" style="background-image: url('{$WebApplicationBaseURL}images/icons/icon_common.png');" />

the hit icon overlay is designed only as an overlay to this common icon

kkrebs (changes requested):
see comment

kkrebs (approved):

PR #713, opened 2022-01-27
MIR-752 add copyrightMD schema to MIR resources

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-752).


kkrebs (changes requested):
Is this MIR specific or should this be part of mycore-mods?

kkrebs (approved):

PR #724, opened 2022-03-11
MIR-1091 - Added description html meta tag to metadata page

xaher:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1091).

### DIFF
--- a/mir-module/src/main/resources/xsl/metadata/mir-citation.xsl
+++ b/mir-module/src/main/resources/xsl/metadata/mir-citation.xsl
@@ -31,36 +31,35 @@
 
     <!-- ==================== Highwire Press Tags, Dublin Core as Meta Tags and SEO meta tags ==================== -->
     <citation_meta>
-          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="dc-meta"/>
-          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="highwire" />
-          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="seo-meta" />
-
+      <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="dc-meta"/>
+      <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="highwire" />
+      <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="seo-meta" />
     </citation_meta>
 
-        <xsl:variable name="piServiceInformation" select="piUtil:getPIServiceInformation(mycoreobject/@ID)" />
-        <xsl:variable name="mods" select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" />
+    <xsl:variable name="piServiceInformation" select="piUtil:getPIServiceInformation(mycoreobject/@ID)" />
+    <xsl:variable name="mods" select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" />
 
-        <div id="mir-citation">
-          <xsl:if test="$MIR.shariff = 'show'">
+    <div id="mir-citation">
+      <xsl:if test="$MIR.shariff = 'show'">
 
-            <xsl:variable name="modsTitle">
-              <xsl:apply-templates select="$mods" mode="title" />
-            </xsl:variable>
-            <xsl:variable name="shariffURL">
-              <xsl:choose>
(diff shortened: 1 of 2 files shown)


mir-module/src/main/resources/xsl/metadata/mir-citation.xsl
```diff
@@ -28,36 +29,38 @@
...
-      <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="highwire" />
+          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="dc-meta"/>
+          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="highwire" />
+          <xsl:apply-templates select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" mode="seo-meta" />
+
     </citation_meta>
 
-    <xsl:variable name="piServiceInformation" select="piUtil:getPIServiceInformation(mycoreobject/@ID)" />
-    <xsl:variable name="mods" select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods" />
+        <xsl:variable name="piServiceInformation" select="piUtil:getPIServiceInformation(mycoreobject/@ID)" />
```
kkrebs, line 40: please use mycore formatting settings for xml/xsl (only 2 spaces instead of 4) to prevent these diffs

mir-module/src/main/resources/xsl/mods-seo-meta.xsl
```diff
@@ -0,0 +1,28 @@
...
+                xmlns:xalan="http://xml.apache.org/xalan"
+                exclude-result-prefixes="xsl mods xlink xalan">
+
+    <xsl:include href="mods-utils.xsl"/>
+
+    <xsl:template match="mods:mods" mode="seo-meta">
+        <xsl:apply-templates select="mods:abstract|mods:abstract/@xlink:href" mode="seo"/>
+    </xsl:template>
+
+    <xsl:template match="mods:abstract[not(@altFormat)] | mods:tableOfContents[not(@altFormat)] | mods:note" mode="seo">
```
kkrebs, line 16: - mods:note should not match here
- we should add a check for abstract in current lang, including fallback if abstract in current lang is not available

kkrebs (changes requested):

mir-module/src/main/resources/xsl/mods-seo-meta.xsl
```diff
@@ -0,0 +1,32 @@
...
+    <xsl:template match="mods:mods" mode="seo-meta">
+        <xsl:apply-templates select="mods:abstract|mods:abstract/@xlink:href" mode="seo"/>
+    </xsl:template>
+
+    <xsl:template match="mods:abstract[not(@altFormat)] | mods:tableOfContents[not(@altFormat)]" mode="seo">
+        <xsl:if test="@xml:lang = $CurrentLang" >
+            <meta>
+                <xsl:attribute name="name">Description</xsl:attribute>
+                <xsl:attribute name="content">
+                    <xsl:value-of select="concat(substring(.,0,141), '...')" />
```
toKrause, line 23: A few quick points:

- I think a "horizontal ellipsis" (https://www.fileformat.info/info/unicode/char/2026/index.htm) should be used rather then three dots.
- Out of interest: Where does the target length of 140 characters come from? Is this common SEO-knowledge? Maybe add a comment.
- I think `substring(.,1,140)` would be clearer here. String indexes are 1-based in XSLT.
- The ellipsing should only be done if the processed abstract has at least that many characters.

mir-module/src/main/resources/xsl/mods-seo-meta.xsl
```diff
@@ -0,0 +1,41 @@
...
+    <xsl:template match="mods:abstract[not(@altFormat)] | mods:tableOfContents[not(@altFormat)]" mode="seo">
+        <xsl:if test="@xml:lang = $CurrentLang" >
+            <meta>
+                <xsl:attribute name="name">description</xsl:attribute>
+                <xsl:attribute name="content">
+                    <xsl:choose>
+                        <!-- SEO descriptions are limited to about 155 characters for desktop search and 120 characters
+                             for mobile search. -->
+                        <xsl:when test="string-length(.)>=140">
+                            <xsl:value-of select="concat(substring(.,1,140), '&amp;hellip;')" />
```
toKrause, line 27: Minor nitpick: The ellipsing is only relevant if the processed abstract actually more then 140 character, so a `>` comparison should be used.

Now that I've looked into this, maybe it would be better to use `mcrxml:shortenText` which appears to be made for exactly a situation like this. It makes the length comparison and picks _nicer_ a cutting by searching for the nearest whitespace.

kkrebs (approved):

PR #734, opened 2022-03-22
MIR-1125 Fix missing, unfitting and inconsistent form validation styling

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1125).

### DIFF
--- a/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/bootstrap4fix.scss
+++ b/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/bootstrap4fix.scss
@@ -32,21 +32,22 @@
 }
 .mcr-invalid {
   label {
-    color: red;
+    color: $danger;
   }
   input,
-  select {
-    border: 1px solid red;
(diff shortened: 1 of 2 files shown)


mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/bootstrap4fix.scss
```diff
@@ -32,21 +32,22 @@
 }
 .mcr-invalid {
   label {
-    color: red;
+    color: $danger;
   }
   input,
-  select {
-    border: 1px solid red;
-    padding-right: calc((1em + .75rem) * 3 / 4 + 1.75rem);
```
kkrebs, line 40: is it a typo, that padding-right is now missing?

toKrause: As discussed in the ticket, that rule doesn't appear to affect anything. If it does, and I just didn't see it, that rule should probably also be present in the corresponding rule for `.mcr-valid`. Either way, dist is an inconsistency that should be addressed.

PR #743, opened 2022-04-01
MIR-1133 - add message property for title

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1133).


kkrebs:
already solved in MIR-1134

PR #745, opened 2022-04-01
MIR-1135 Incorrect I18N keys

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1135).


mir-module/src/main/resources/META-INF/resources/editor/editor-dynamic.xed
```diff
@@ -29,10 +29,10 @@
         <xed:param name="series" default="no" />
 
         <h1>
-          <xed:output i18n="mir.form.editor.dynamic.header" />
+          <xed:output i18n="editor.dynamic.header" />
```
kkrebs, line 32: please leave mir.form prefix here and add it in message.property to fix this

kkrebs (changes requested):

kkrebs (approved):

PR #761, opened 2022-04-28
MIR-1143 - XMetaDissPlus transfer url should only contain ASCII characters

yagee-de:
should only contain ASCII characters according to DNB (undocumented)

[Link to jira](https://mycore.atlassian.net/browse/MIR-1143).


fluetze (approved):

fluetze:
Suggest to use a separate variable:
            <xsl:variable name="encodedPath" select="mcrxsl:encodeURIPath(mcrxsl:decodeURIPath($filePath), true())" />
            <xsl:value-of select="concat($ServletsBaseURL,'MCRFileNodeServlet/',$derId,$encodedPath)" />
Avoids long line

PR #810, opened 2022-10-06
MIR-1175 repair MIR-Wizard

sebhofmann:
- use MCRConfigurationDirSetup.loadExternalLibs() instead of reflection to load new driver lib
- fixed generated persistence.xml
- updated README.md to use tomcat instead of jetty

[Link to jira](https://mycore.atlassian.net/browse/MIR-1175).


Mewel (approved):
Works as expected

PR #827, opened 2022-12-01
MIR-1186 allow additional authority attributes for role term element

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1186).


mir-module/src/main/resources/xsl/editor/xeditor2mods.xsl
```diff
@@ -215,14 +215,19 @@
           <xsl:copy-of select="node()[name()!='mods:namePart']" />
         </xsl:when>
         <xsl:otherwise>
-          <xsl:copy-of select="node()" />
+          <xsl:copy-of select="node()[name()!='mods:role']" />
```
toKrause, line 218: It should be sufficient to just switch from `copy-of` to `apply-templates` (for all nodes), i.e.

```
<xsl:choose>
  <xsl:when test="@simpleEditor">
    <xsl:apply-templates select="node()[name()!='mods:namePart']"/>
  </xsl:when>
  <xsl:otherwise>
    <xsl:apply-templates/>
  </xsl:otherwise>
</xsl:choose>
```
The special treatment `<xsl:if test="mods:role/mods:roleTerm">` is then unnecessary. I've tested this both with a simple editor (i.e. `editor-dynamic.xed/`) and a normal editor (i.e. `editor-admins.xed`). I didn't notice any negatice or unintended effects and this should provide more flexibility for future modifications.

toKrause (changes requested):

mir-module/src/main/resources/xsl/editor/xeditor2mods.xsl
```diff
@@ -215,7 +215,7 @@
           <xsl:copy-of select="node()[name()!='mods:namePart']" />
```
toKrause, line 215: I would have switched to `apply-templates` here too. 

Since `copynodes.xsl` is included, the default behavior is to copy anyway. Switching to `apply-templates` allows modifications in the future. I see no reason, not to allow this in both cases.

This inclusion `copynodes.xsl` predates this template. I sincerely believe that the use of `copy-of` in this template was just an oversight. An oversight that should be corrected. The issue that originally lead to to this PR is just a symptom of this root cause.

toKrause (changes requested):

toKrause (approved):

kkrebs (approved):

PR #832, opened 2022-12-16
MIR-1189 improve docker containers

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1189).

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,5 @@
+NAME=mir
+
 DB_PORT=8293
 DB_USER=pguser
 DB_PASSWORD=password
@@ -8,6 +10,9 @@ MIR_HTTP=8291
 MIR_AJP=8292
 MIR_HOME=./docker/mir-home/
 MIR_DATA=./docker/mir-data/
+MIR_LOGS=./docker/mir-logs/
+MIR_XMX=1024m
+MIR_XMS=1024m
 
 SOLR_HTTP=8290
 SOLR_DATA=./docker/solr-data/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,32 @@
 FROM tomcat:jdk17-temurin-focal
-EXPOSE 8080
-EXPOSE 8009
+RUN groupadd -r mcr -g 501 && \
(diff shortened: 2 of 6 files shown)


os-cau (commented):
It seems to me that the docker-entrypoint.sh script in the container will be owned by the user mcr, but executed as root when the container starts. This would mean that the mcr user is effectively root, because he could add arbitrary commands to the script which would be executed as root on restart.

The owner of the script should be root, if it is executed as root. It might also be possible to move the entire root-section from docker-entrypoint.sh to the Dockerfile and set mcr as the executing user there via the USER statement. This would remove root from the picture entirely.

sebhofmann:
> It seems to me that the docker-entrypoint.sh script in the container will be owned by the user mcr, but executed as root when the container starts. This would mean that the mcr user is effectively root, because he could add arbitrary commands to the script which would be executed as root on restart.

Sound good, i will change that. 
Change  `COPY --chown=mcr:mcr docker-entrypoint.sh /usr/local/bin/mir.sh` to  `COPY --chown=root:root docker-entrypoint.sh /usr/local/bin/mir.sh` and adding chmod 555 /usr/local/bin/mir.sh to the next run command should be save, right?

> The owner of the script should be root, if it is executed as root. It might also be possible to move the entire root-section from docker-entrypoint.sh to the Dockerfile and set mcr as the executing user there via the USER statement. This would remove root from the picture entirely.

The root user is used to change the rights of the mount path. The mount path content is not present at time the Dockerfile runs, so i can not adjust the rights. Its not even simple to change that rights on the host in Ubuntu rootless environment, because ubuntu maps the root in the container to the user id, but every other id in the container is mapped to a id which is generated from /etc/subuid (as offset) and the user id in the docker container. So you would need to know the internal id of the user in the container to prepare the mounted folder. So for convenient reasons, i chose to change the rights at container startup, which only has the disadvantage that you can not edit the files outside the container without extra permission.

os-cau:
Yes, the modifications you propose are exactly what I had in mind. They should prevent this from happening.

kkrebs (approved):

PR #851, opened 2023-03-13
MIR-1208 Fixed font size in '.dateExtended_box .mir-fieldset-legend, …

Possommi:
….personExtended_box .mir-fieldset-legend' (forms.scss) and typo in label 'mir.date.issued.range'

[Link to jira](https://mycore.atlassian.net/browse/MIR-1208).


mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/forms.scss
```diff
@@ -139,7 +139,7 @@
 }
 
 .dateExtended_box .mir-fieldset-legend, .personExtended_box .mir-fieldset-legend {
-    font-size: 14px;
+    font-size: 1.1rem;
```
kkrebs, line 142: we should never set font-size directly in such cases, use variables instead
perhaps $font-size-base is the right one here?

kkrebs (changes requested):

mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/forms.scss
Possommi: I will give it a try maybe $font-size-base will be to small.

kkrebs: In this case we should look which variable fits or set this value relative to $font-size-base

kkrebs (approved):

PR #854, opened 2023-03-15
MIR-1209 fix new person editor and add command to fix wrong ids

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1209).

### DIFF
--- a/mir-module/src/main/resources/xsl/editor/xeditor2mods.xsl
+++ b/mir-module/src/main/resources/xsl/editor/xeditor2mods.xsl
@@ -144,9 +144,9 @@
         </xsl:otherwise>
       </xsl:choose>
     </xsl:variable>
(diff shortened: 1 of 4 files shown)


mir-module/src/main/resources/xsl/editor/xeditor2mods.xsl
```diff
@@ -145,6 +145,12 @@
       </xsl:choose>
     </xsl:variable>
     <mods:nameIdentifier type="{$type}" typeURI="{$uri}">
+      <xsl:if test="string-length($uri) &gt; 0">
+        <xsl:attribute name="valueURI">
```
kkrebs, line 149: The right attribute here is typeURI, not valueURI

kkrebs (changes requested):

PR #859, opened 2023-03-29
MIR-1213 update jats2mods.xsl to xslt 3 and add thulb fixes

eckstueck:
- update jats2mods.xsl to xslt 3
- use oa_nlz only as default
- fixed solr query for ISSN
- fixed empty contrib-id
- changed mods:dateIssued to w3cdtf
- improved license parsing

[Link to jira](https://mycore.atlassian.net/browse/MIR-1213).


kkrebs (changes requested):
see comment

fluetze (changes requested):
The check <xsl:if test="string-length($orcid)>0"> should not be necessary. This should be adressed by the template commented with "Ignore empty ORCIDs". Please check or fix that template instead.

One could use the same syntax at <xsl:template match="contrib-id">

In the template permissions/license,
       <xsl:otherwise>
          <xsl:value-of select="''"/>
        </xsl:otherwise>
is obsolete and can be removed

kkrebs (approved):

PR #861, opened 2023-05-04
MCR-2856 changes required by api changes in mycore

sebhofmann:
Requires https://github.com/MyCoRe-Org/mycore/pull/1835


Mewel (approved):
Looks good!

PR #865, opened 2023-05-22
MIR-1214 encode query only not the whole url

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1214).


rsteph-de (approved):
works as discussed before

PR #896, opened 2023-08-17
MIR-1238 Forced friends aren't real friends

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1238).

PR #903, opened 2023-10-10
MIR-1244 Remove default OAI friends configuration, related to MCR-2509

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1244).


toKrause:
This has already been dealt with in [MIR-1238](https://github.com/MyCoRe-Org/mir/pull/896); in a nicer way (comment out instead of delete, switch to HTTPS).

I suggest closing this PR and maybe cherry-picking MIR-1238 into 2022.06.x.

PR #918, opened 2023-11-13
Issues/mir 1256 support time in mods date issued

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1256).


mir-module/src/main/resources/META-INF/resources/js/mir/xeditor-form.js
```diff
@@ -284,11 +317,16 @@ $(document).ready(function() {
      // Enables the datetimepicker
      if (jQuery.fn.datepicker) {
          $('.datetimepicker').find('input').each( function(index, elm){
-             pickDatePickerFormatAndAdd(elm);
+             if ($(elm).val().includes("T")) {
```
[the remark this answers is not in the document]
Antoniafriedrich: Thank you for bringing this to my attention. The code you mentioned is located within the "$('.datetimepicker').find('input').each( function(index, elm){...}" block, and only executed when editing a document that already contains a date or datetime value in one of its input fields.

It is not intended to interfere with the creation of a new document but rather comes into play during the editing process. The purpose of this code is to recognize fields with datetime values and automatically select the "Date and Time" format upon loading the editor, instead of the standard "date" format. 

To use the timepicker when creating a document, you can click the grey arrow on the right side on the input field and select "Date and Time" to have the timepicker included. By default it is only set to "date" so you can not pick a time.
I hope this clears things up. If there are further questions please feel free to let me know.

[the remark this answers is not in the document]
toKrause: Sorry, but i need to disagree here.

1. A technicality: in your first paragraph, your write "The code you mentioned is [...] only executed when editing a document that already contains a date or datetime value in one of its input fields." That is not correct. `$('.datetimepicker').find('input').each(...);` specifically means that the code is executed for every input field marked with class `datetimepicker` in every form. Whether or not they already contain a datetime value is irrelevant. 
2. More importantly: you say  that "[i]t is not intended to interfere with the creation of a new document". That might be true for _the specific use case you have in mind_, but this component is a general component that is and will be used for many other use cases. The component should therefore behave in a consistent way.
    - As a user I'd expect the same capability (to select a time value in addition to a date value) independent of whether I create a new document or edit an existing one.
    - From a metadata point of view: if it makes sense for a form field to contain date and time information during editing, I'd suggest that this also makes sense during creation of a new document. 
3. Still more importantly and as alluded to in my previous comment: this component is already in use. Up until now it was only used to select date information. Supporting time information as well fundamentally changes the nature of this component. That is certainly a useful improvement, but some form of backwards compatibility to support the existing uses (where only date information is required and time information is neither needed nor wanted) **is a must**. If that requires a migration step, that would be fine for me, but the target branch needs to be changed then.

Lastly, I've checked out this PR in conjunction with the current 2023.06.1-SNAPSHOT version of MyCoRe, and the "grey arrow on the right side on the input field" you talk about isn't shown to me.

Example (form for a dissertation with two fields that have the class `datetimepicker`):
image

mir-module/src/main/resources/config/mir/mycore.properties
```diff
@@ -590,7 +590,7 @@ MCR.Jersey.Resource.Packages=%MCR.Jersey.Resource.Packages%,org.mycore.mir.stati
 ##############################################################################
 # xEditor                                                                    #
 ##############################################################################
-
+MIR.Editor.AddTimeWidget=true
```
toKrause, line 593: This might go too far in the other direction. As I said before, having the new feature is a nice and welcome improvement. With this addition, system maintainers have to choose: only date fields or only date-and-time-fields. Why not allow them to make a case-by-case decision as suggested in my original comment?

toKrause: Also, when targeting the LTS branch, the default value should be `false`, so that the established behavior is not changed.

toKrause:
It's been a few days since I last looked at this PR.

Please remind me, with the current implementation, how is the per-field-configuration made, to set whether it is a date-and-time-field or just a date-field?

Antoniafriedrich:
In editor-includes.xed, you can configure two attributes to control the behavior of date fields:

**showDateTimeOption**: Set this to true to include the time option in date fields. This mainly influences whether you can choose this option if you have a dropdown with many options such as Date, Daterange and Datetime.

**startsWithDateTime**: Set this to true to make the date-time option the default when the page with the date field is loaded.

These attributes can be added to the textfields of templates inside editor-includes.xed. If these attributes are left empty or set to false, the standard behavior of the date fields remains unchanged.
It should be possible to create a new template with these attributes and add it to your MyCoReWebPage.

The logic for that is mostly handled in mir2xeditor. 

I have added the attributes as an example in one template, set to false, so the behavior should not change unless changed to true.

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -129,7 +130,10 @@
         <xed:output i18n="{@label}" />
       </label>
       <div class="col-md-6 {@class}" data-type="{@type}">
-        <xsl:call-template name="mir-dateRange"/>
+        <xsl:call-template name="mir-dateRange">
+          <xsl:with-param name="showDateTimeOption" select="@showDateTimeOption"/>
```
toKrause, line 134: There is a typing issue here. Values for `showDateTimeOption` and `showDateTimeOption` from `editor-includes.xed` are attribute values and therefore of type "String", ...

... they are passed through as parameters here in the `mir:dateRange` template ...

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -141,11 +145,35 @@
...
     <div class="{@class}" data-type="{@type}">
-      <xsl:call-template name="mir-dateRange"/>
+      <xsl:call-template name="mir-dateRange">
+        <xsl:with-param name="showDateTimeOption" select="@showDateTimeOption"/>
+      </xsl:call-template>
     </div>
   </xsl:template>
 
   <xsl:template name="mir-dateRange">
+    <xsl:param name="showDateTimeOption" select="false" />
```
toKrause, line 155: ... but the default values here are of type "Element" ...

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -141,11 +145,35 @@
...
+          <xsl:text>startsWithDatetime</xsl:text>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:text></xsl:text>
+        </xsl:otherwise>
+      </xsl:choose>
+    </xsl:variable>
+    <xsl:variable name="dateRangeWithTime">
+      <xsl:choose>
+        <xsl:when test="$showDateTimeOption='true'">
```
toKrause, line 169: ... but compared to a string literal ...

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -165,32 +193,37 @@
...
       <div class="date-simple {$hiddenclasssimple} input-group mb-1">
         <xed:bind xpath="{$xpathSimple}">
-          <input type="text" class="form-control" autocomplete="off">
+          <input type="text" class="form-control {$timeClass}" autocomplete="off">
             <xsl:copy-of select="@placeholder" />
           </input>
         </xed:bind>
-        <xsl:call-template name="date-selectFormat"/>
+        <xsl:call-template name="date-selectFormat">
+          <xsl:with-param name="showDateTimeOption" select="$showDateTimeOption"/>
```
toKrause, line 201: ... and passed through to the `date-selectFormat` template ...

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -165,32 +193,37 @@
...
+          <xsl:call-template name="date-selectFormat">
+            <xsl:with-param name="showDateTimeOption" select="$showDateTimeOption"/>
+          </xsl:call-template>
         </xsl:if>
       </div>
     </div>
   </xsl:template>
 
   <xsl:template name="date-selectFormat">
+    <xsl:param name="showDateTimeOption" select="false" />
```
toKrause, line 226: ... where the default value is again of type "Element" ...

mir-module/src/main/resources/xsl/editor/mir2xeditor.xsl
```diff
@@ -204,6 +237,13 @@
             <xsl:value-of select="i18n:translate('mir.date.period')" />
           </a>
         </li>
+        <xsl:if test="$showDateTimeOption='true'">
```
toKrause, line 240: ... but compared to a string literal.

I think, the default values should be string literals as well.

toKrause:
I've noticed a potential problem (see comments).

I've also tried out your PR extensively with my application and an unmodified MIR (current version of MyCoRe- and MIR-`main`-branch respectively) and was so far unable to even see the UI for entering a time value, let alone check, if everything works consistently. 

Please demonstrate the feature during the next web conference.

toKrause (approved):

PR #925, opened 2023-11-20
MIR-1258 Set left attribute in .hit_icon_overlay to 16px

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1258).

### DIFF
--- a/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_results.scss
+++ b/mir-layout/src/main/resources/META-INF/resources/mir-layout/scss/common/mir_results.scss
@@ -226,7 +226,7 @@ $filter: #666 !default;
       .hit_icon_overlay {
         position: absolute;
         top: 20px;
-        left: -8px;
+        left: 16px;
         width: 50%;
       }
       .hit_tnd_container {


toKrause (changes requested):
In my repository (which uses slightly larger icons then standard MIR):

Currently (definitely broken):
image

With proposed changes (still not correct):
image

Should be dynamically centered via CSS, for example with:
```
  left: 0;
  right: 0;
  margin-left: auto;
  margin-right: auto;
```
image

Or, preferably, taking the visual imbalance of the icons into account:
```
  left: -8%;
  right: 0;
  margin-left: auto;
  margin-right: auto;
```
image

Appears to be working for different sizes: 
image

sebhofmann:
I think the current position of the arrow is intentional.

toKrause:
> I think the current position of the arrow is intentional.

very weird, if this is intentional

image

sebhofmann:
I think it looks weird too, i tried to fix it, when i migrated the icons to IIIF, but i was told the arrow should line up with the border and it is intentional. 
Bildschirmfoto 2023-11-20 um 15 31 03

Possommi:
> that looks ridiculous, imho

I am quite new to MIR and to be honest when I first saw it I thought that the layout must be broken.

PR #901, opened 2023-10-02
MIR-1241 Response mir facets should be dynamic

levshyn:
Facets can be added to the address bar

[Link to jira](https://mycore.atlassian.net/browse/MIR-1241).

PR #938, opened 2023-12-04
MIR-1262 Errors in MIR-1251

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1262).


Possommi:
already been fixed here https://github.com/MyCoRe-Org/mir/pull/901

toKrause:
Both parts have indeed been fixed in #901. This PR should only be merged as a quick fix, if #901 needs to be discussed and is not merged shortly.

PR #966, opened 2024-02-28
MCR-3031 fixed compile error and the wizard now uses properties instead of persistence.xml

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3031).


mir-wizard/src/main/java/org/mycore/mir/wizard/command/MIRWizardGenerateProperties.java
```diff
@@ -65,7 +65,9 @@ public void doExecute() {
                     MCRConfiguration2.set(name, value);
                 }
             }
-            mcrProps.store(new FileOutputStream(file), "Generated by MIR Wizard.");
+            try(FileOutputStream out = new FileOutputStream(file)){
```
toKrause, line 68: I think, as a precaution,  `MIRWizardGenerateProperties` should also append to the selected `mycore.properties` file, rather then overwrite it completely. Otherwise, a reordering of `install.xml` could lead to missing properties.

toKrause (changes requested):

mir-wizard/src/main/java/org/mycore/mir/wizard/command/MIRWizardGenerateProperties.java
sebhofmann: done

toKrause (approved):

PR #969, opened 2024-03-04
MIR-1280 Links from details page to name identifier based name search…

kkrebs:
… not working (#959)

* MIR-1280 omit encoding the parameter seperating ampersand

* MIR-1280 encode entire parameter value

[Link to jira](https://mycore.atlassian.net/browse/MIR-1280).


kkrebs:
@toKrause : we need this bugfix also in 2022.06.x

PR #971, opened 2024-03-08
MIR-1285 Display realname rather than uid of user after login

Possommi:
[MIR-1285](https://mycore.atlassian.net/browse/MIR-1285).

[MIR-1285]: https://mycore.atlassian.net/browse/MIR-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mir-it/src/test/java/org/mycore/mir/it/controller/MIRUserController.java
+++ b/mir-it/src/test/java/org/mycore/mir/it/controller/MIRUserController.java
@@ -19,8 +19,8 @@ import org.openqa.selenium.support.ui.ExpectedConditions;
 public class MIRUserController {
 
     public static final String ADMIN_PASSWD = "alleswirdgut";
+
     public static final String ADMIN_LOGIN = "administrator";
-    public static final String ADMIN_REALNAME = "Superuser";
 
     String baseURL;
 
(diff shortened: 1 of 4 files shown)


toKrause (approved):

toKrause (changes requested):
This needs coordination with [MIR-1288](https://github.com/MyCoRe-Org/mir/pull/974)

This implementation uses the name, if available, in the UI. MIR-1288 still uses the user ID.

toKrause (approved):

fluetze:
A user's real name is typically much wider than the user ID. So this change will affect the layout of MIR instances nad might be unwanted. Or is there any customization option?

toKrause:
Since this is a behavior change that targets the LTS branch, the change should technically
be optional, i.e. configurable. But this specific change is so undramatic that I wouldn't insist on that.

Possommi:
I am with @toKrause on the matter.

From user perspective its more friendly to read the name rather than the uid. In particular when a user is signed in via SSO where the username could be anything (uid, scroped attribute or even some sort of uuid).

Possommi:
I will introduce a new property making the behaviour configurable

Possommi:
> A user's real name is typically much wider than the user ID. So this change will affect the layout of MIR instances nad might be unwanted. Or is there any customization option?

@fluetze I added a property and the default behaviour remains unchanged

PR #976, opened 2024-03-12
MIR-1291 Submitters should be enabled to see the history of their records

Possommi:
[MIR-1291](https://mycore.atlassian.net/browse/MIR-1291)

[MIR-1291]: https://mycore.atlassian.net/browse/MIR-1291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

requires https://github.com/MyCoRe-Org/mycore/pull/2110

### DIFF
--- a/mir-module/src/main/resources/config/mir/mycore.properties
+++ b/mir-module/src/main/resources/config/mir/mycore.properties
@@ -66,6 +66,7 @@ MCR.Access.Strategy.Class=org.mycore.mir.authorization.MIRStrategy
 # comma separated classification IDs that are relevant for ACL checks
 MIR.Access.Strategy.Classifications=mir_access
 MIR.Strategy.EditPIRoles=admin
+#MCR.Access.Strategy.CreatorAnyStatePermissions=read-history
 MCR.Access.Strategy.CreatorRole=submitter
 MCR.Access.Strategy.SubmittedCategory=state:submitted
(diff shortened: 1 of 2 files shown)


sebhofmann:
We have a acl-check for reading history of an object, maybe we should use that instead. We have 'view-history' to see the history and 'read-history' to read the actual objects.

Possommi:
> We have a acl-check for reading history of an object, maybe we should use that instead. We have 'view-history' to see the history and 'read-history' to read the actual objects.

but how do we allow access for the original submitter of the document without adding that submitter to the rule itself

sebhofmann:
> > We have a acl-check for reading history of an object, maybe we should use that instead. We have 'view-history' to see the history and 'read-history' to read the actual objects.
> 
> but how do we allow access for the original submitter of the document without adding that submitter to the rule itself

With rules.xml its should be easy, but with the default MIRStrategy its not possible. The org.mycore.access.strategies.MCRCreatorRuleStrategy which is used by MIRStrategy would only work if the document is in the state submitted. Maybe the MCRCreatorRuleStrategy could be extended by MCR.Access.Strategy.AllPermissions=read-history. I would really prefer to have this controlled by the ACL.

sebhofmann (approved):

PR #978, opened 2024-03-13
MIR-1290 Open link to ORCID in new window/tab in name search

Possommi:
[MIR-1290](https://mycore.atlassian.net/browse/MIR-1290).

[MIR-1290]: https://mycore.atlassian.net/browse/MIR-1290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mir-webapp/src/main/vue/name-search/src/components/IdentifierDisplay.vue
+++ b/mir-webapp/src/main/vue/name-search/src/components/IdentifierDisplay.vue
@@ -17,7 +17,7 @@
   -->
 
 <template>
-  <span><span class="identifier-key">{{ type }}</span><a class="identifier-value" :href="url">{{ value }}</a> </span>
+  <span><span class="identifier-key">{{ type }}</span><a class="identifier-value" :href="url" target="_blank">{{ value }}</a> </span>
 </template>
 
 <script setup lang="ts">


mir-webapp/src/main/vue/name-search/src/components/IdentifierDisplay.vue
```diff
@@ -17,7 +17,7 @@
   -->
 
 <template>
-  <span><span class="identifier-key">{{ type }}</span><a class="identifier-value" :href="url">{{ value }}</a> </span>
+  <span><span class="identifier-key">{{ type }}</span><a class="identifier-value" :href="url" target="_blank">{{ value }}</a> </span>
```
kkrebs, line 20: hmm ...

So far I have tried to avoid links in MIR that are opened hard in a new tab. But I realise that there is a need ("The [question of ethics](https://www.searchenginejournal.com/blank-link-attribute/435883/) has largely been set aside nowadays."). But I think it would be good if we could define a kind of best practice for this in MIR.

toKrause: Yes, as far as i remember, we decided to not use `target="_blank"`; users should be able to choose the behavior they want.

toKrause: Should we decide to change that, then _all_ links to external pages should forcefully be opened in a new browser tab.

kkrebs: Yes, I've also thought about that in the meantime. That's what we've decided and we shouldn't deviate from this design decision now. If we do, then it should be a consistent implementation everywhere.

PR #980, opened 2024-03-15
MIR-1288 add acl for displaying submitted document

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1288).

### DIFF
--- a/mir-module/src/main/resources/rules/rules-webpage.xml
+++ b/mir-module/src/main/resources/rules/rules-webpage.xml
@@ -70,6 +70,14 @@
                 <role>admin</role>
                 <role>editor</role>
             </or>
+            <or>
+                <not>
+                    <regex>webpage:/servlets/solr/select\?q=createdby:{CurrentUser}&amp;fq=objectType:mods.*</regex>
+                </not>
+                <role>admin</role>
+                <role>editor</role>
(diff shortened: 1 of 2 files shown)


kkrebs (approved):

kkrebs (changes requested):
please take a look at MIREditorACLTest - there are 6 errors

seems to be a problem with `{CurrentUser}` notation

mir-module/src/main/resources/rules/rules-webpage.xml
```diff
@@ -70,6 +70,14 @@
                 <role>admin</role>
                 <role>editor</role>
             </or>
+            <or>
+                <not>
+                    <regex>webpage:/servlets/solr/select\?q=createdby:{CurrentUser}&amp;fq=objectType:mods.*</regex>
```
sebhofmann, line 75: Using plain { and } in a regexp seems wrong

sebhofmann (changes requested):

mir-module/src/main/resources/rules/rules-webpage.xml
yagee-de: Thanks for the hint!

sebhofmann (approved):

PR #987, opened 2024-03-26
MIR-1298 default unapi database have to be k10plus

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1298).


rsteph-de:
There could be a problem the k10plus database SRU interface is not public available.
You need to have an IP from a GBV library.
``` xml
<diag:message>Access to specified database denied</diag:message>
```

sebhofmann (approved):

PR #996, opened 2024-04-17
MIR-1306 sword dependencies using javax

eckstueck:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1306).

### DIFF
--- a/mir-webapp/pom.xml
+++ b/mir-webapp/pom.xml
@@ -18,6 +18,8 @@
     <solr.coreName>mir</solr.coreName>
     <solr.data.dir>${mir.solr.data.dir}</solr.data.dir>
     <solr.home>${mir.solr.home}</solr.home>
+    <abdera.version>1.1.3</abdera.version>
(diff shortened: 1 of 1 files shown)


rsteph-de (changes requested):
see my comment in JIRA ticket

yagee-de (approved):

yagee-de:
> > > see my comment in JIRA ticket
> > 
> > 
> > If this PR fixes the problem, it's looking like an easy fix to be applied in other applications as well. You're right, that this is fundamentally a mycore-sword problem that should be fixed there. BUT I would rather see a working workaround merged soon in MIR than to wait for a harder to implement general fix in mycore-sword.
> 
> Why is it harder to fix this in mycore? It think would be almost the same pr, just with changes in mycore-sword, no?

In mycore is no WAR-package just JAR-packages. So either you have to provide a fat-jar or fork 5 dependencies and provide a fix.

rsteph-de (approved):

PR #1031, opened 2024-07-25
MIR-1336 Display realname of creator/modifier in mir-admin-box

Possommi:
[MIR-1336](https://mycore.atlassian.net/browse/MIR-1336).

[MIR-1336]: https://mycore.atlassian.net/browse/MIR-1336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mir-module/src/main/resources/xsl/metadata/mir-admindata-box.xsl
+++ b/mir-module/src/main/resources/xsl/metadata/mir-admindata-box.xsl
@@ -5,6 +5,7 @@
                 xmlns:mcrxsl="xalan://org.mycore.common.xml.MCRXMLFunctions"
                 exclude-result-prefixes="i18n mods xlink mcrxsl">
   <xsl:import href="xslImport:modsmeta:metadata/mir-admindata-box.xsl"/>
+  <xsl:param name="WebApplicationBaseURL"/>
 
   <xsl:template match="/">
     <xsl:variable name="ID" select="/mycoreobject/@ID"/>
@@ -30,9 +31,10 @@
               <xsl:with-param select="mycoreobject/service/servdates/servdate[@type='createdate']" name="nodes"/>
               <xsl:with-param select="i18n:translate('metaData.createdAt')" name="label"/>
             </xsl:call-template>
-            <xsl:call-template name="printMetaDate">
-              <xsl:with-param select="mycoreobject/service/servflags/servflag[@type='createdby']" name="nodes"/>
-              <xsl:with-param select="i18n:translate('mir.metaData.detailBox.by')" name="label"/>
+
+            <xsl:call-template name="print-user-info">
+              <xsl:with-param name="user" select="document(concat('notnull:user:', mycoreobject/service/servflags/servflag[@type='createdby']))"/>
+              <xsl:with-param name="label" select="i18n:translate('mir.metaData.detailBox.by')"/>
             </xsl:call-template>
             <xsl:for-each select="mycoreobject/metadata/def.modsContainer/modsContainer/mods:mods/mods:note">
               <xsl:variable name="noteType">
@@ -65,9 +67,9 @@
               <xsl:with-param select="mycoreobject/service/servdates/servdate[@type='modifydate']" name="nodes"/>
               <xsl:with-param select="i18n:translate('metaData.lastChanged')" name="label"/>
             </xsl:call-template>
-            <xsl:call-template name="printMetaDate">
-              <xsl:with-param select="mycoreobject/service/servflags/servflag[@type='modifiedby']" name="nodes"/>
-              <xsl:with-param select="i18n:translate('mir.metaData.detailBox.by')" name="label"/>
+            <xsl:call-template name="print-user-info">
+              <xsl:with-param name="user" select="document(concat('notnull:user:', mycoreobject/service/servflags/servflag[@type='modifiedby']))"/>
+              <xsl:with-param name="label" select="i18n:translate('mir.metaData.detailBox.by')"/>
             </xsl:call-template>
             <!--*** MyCoRe-ID and intern ID *************************** -->
             <tr>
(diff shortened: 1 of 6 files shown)


mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl
```diff
@@ -30,10 +30,12 @@
             <xsl:with-param select="mycoreobject/service/servdates/servdate[@type='createdate']" name="nodes"/>
             <xsl:with-param select="i18n:translate('metaData.createdAt')" name="label"/>
           </xsl:call-template>
-          <xsl:call-template name="printMetaDate">
-            <xsl:with-param select="mycoreobject/service/servflags/servflag[@type='createdby']" name="nodes"/>
-            <xsl:with-param select="i18n:translate('metaData.createdby')" name="label"/>
+
```
toKrause, line 33: Ideally, both representations (username and real name) are displayed, rather than just one or the other.

mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl
```diff
@@ -129,4 +133,30 @@
...
+    <xsl:param name="label"/>
+
+    <tr>
+      <td class="metaname">
+        <xsl:value-of select="concat($label, ':')"/>
+      </td>
+      <td class="metavalue">
+        <xsl:choose>
+          <xsl:when test="$user/user/realName">
+            <xsl:attribute name="title">
```
toKrause, line 148: This attribute does not survive the layouting process.

mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl
```diff
@@ -129,4 +133,30 @@
...
+            </xsl:attribute>
+            <xsl:value-of select="$user/user/realName"/>
+          </xsl:when>
+          <xsl:otherwise>
+            <xsl:value-of select="$user/user/@name"/>
+          </xsl:otherwise>
+        </xsl:choose>
+      </td>
+    </tr>
+
```
toKrause, line 159: Why reinvent the layout rendering instead of deferring to the existing `printMetaDate` template?

Suggestion:

```
  <xsl:template name="print-user-info">
    <xsl:param name="user"/>
    <xsl:param name="label"/>

    <xsl:variable name="realName">
      <xsl:choose>
        <xsl:when test="$user/user/realName">
          <xsl:value-of select="concat($user/user/realName,' (',$user/user/@name,')')"/>
        </xsl:when>
        <xsl:otherwise>
          <xsl:value-of select="$user/user/@name"/>
        </xsl:otherwise>
      </xsl:choose>
    </xsl:variable>

    <xsl:call-template name="printMetaDate">
      <xsl:with-param select="exslt:node-set($realName)" name="nodes"/>
      <xsl:with-param select="$label" name="label"/>
    </xsl:call-template>

  </xsl:template>

**Review by toKrause — CHANGES_REQUESTED** [review]

File: mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl (hunk 3 above)
Comment (Possommi) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: But then no tooltip will be displayed.

Comment (toKrause) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: What tooltip? Using the stylesheet from your PR, the following HTML was generated:

```
<dl>
<dt>Publikationsstatus:</dt>
<dd>angelegt</dd>
<dt>Erstellt am:</dt>
<dd>25.07.2024 - 07:54:12</dd>
<dt>Erstellt von:</dt>
<dd>Superuser</dd>
<dt>Letzte Änderung:</dt>
<dd>25.07.2024 - 08:38:04</dd>
<dt>Geändert von:</dt>
<dd>Superuser</dd>
<dt>MyCoRe ID:</dt>
<dd>macau_mods_00000001</dd>
<dt>Version:</dt>
<dd>20<br>
<a style="cursor: pointer" id="historyStarter">(Gesamte Versionsgeschichte)</a>
</dd>
</dl>
```

As I said, the `title`-attribute doesn't appear to survive the layouting process.

Comment (Possommi) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: This is strange since I added a modification in mods-metadata-page.xsl to retain the title attribute

Comment (toKrause) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: Even then, I'd like my approach better, because it is impossible (or at least cumbersome) to copy the username from a tooltip.

Copying the username i a usual step in many administrative tasks, especially when trying to investigate some issues.

Comment (Possommi) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: > Specifically, [this line](https://github.com/MyCoRe-Org/mir/blob/622b47518e4ab8780a9aee783b2cf4e89c7ae570/mir-module/src/main/resources/xslt/metadata/mods-metadata-page.xsl#L231) in `mods-metadata-page.xsl` would have to be changed from `<xsl:copy-of select="node()|*"/>` to `<xsl:copy-of select="node()|*|@*"/>` for the attribute to survive.

this would copy all attributes `@class` included and might infer with existing layouts. I do get your point with copying the username to the clipboard. Maybe linking to the user profile page would be an option?

Comment (toKrause) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: Okay, maybe I didn't integrate your change to `mods-metadata-page.xsl` properly into my test setup (I couldn't just apply the PR because I'm in the middle of something else right now), so please disregard my comments about the attribute not surviving the layout process. My bad. Sorry for the inconvenience.

Comment (toKrause) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: Linking could indeed be fine. I'm not sure right now, is the admin metadata box actually only ever displayed to admins, i.e. to users who would actually be allowed to see the linked user profile page?

Comment (Possommi) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: Usually only admins can see the box but in my repo the creators are allowed to see them as well. Unfortunately the user cannot be determined during display since the content is static.

Comment (toKrause) [reply mir-module/src/main/resources/xsl/metadata/static/mir-admindata-box-static.xsl:159]: If it can not be guaranteed the the admin metadata box will only be displayed to admins, linking to the user profile might not be the best idea.

File: mir-module/src/main/resources/xsl/metadata/mir-admindata-box.xsl [hunk 4]
```diff
@@ -126,4 +128,44 @@
...
+
+      <td class="metavalue">
+        <xsl:if test="$user/user/realName">
+          <xsl:attribute name="title">
+            <xsl:value-of select="concat($user/user/@name, '@', $user/user/@realm)"/>
+          </xsl:attribute>
+        </xsl:if>
+
+        <xsl:choose>
+          <xsl:when test="mcrxsl:isCurrentUserInRole('admin')">
```
toKrause, line 159: I think it would be better to check if the current user ha the `administrate-users` permission.

For example, a repository operator may also have permitted editors to manage users.

toKrause (changes requested):

toKrause (approved):

PR #1037, opened 2024-07-26
MIR-1333 Added placeholder for username input field in new-author.xed

Possommi:
[MIR-1333](https://mycore.atlassian.net/browse/MIR-1333).

### DIFF
--- a/mir-module/src/main/resources/config/mir/messages_en.properties
+++ b/mir-module/src/main/resources/config/mir/messages_en.properties
@@ -864,3 +864,4 @@ subselect.category.cancel = Cancel
 subselect.category.title  = Choose category
 
 user.profile.id.orcid = Your ORCID iD
+mir.register.username.placeholder = maximum of 16 characters, allowed characters are a-z, 0-9 and ._-
--- a/mir-module/src/main/resources/META-INF/resources/authorization/new-author-includes.xed
+++ b/mir-module/src/main/resources/META-INF/resources/authorization/new-author-includes.xed
@@ -7,7 +7,7 @@
 
   <xed:template id="userAccount">
     <mcruser:template name="textInput" xpath="@name" id="userName" i18n="component.user2.admin.userAccount" i18n.error="component.user2.admin.inputhints.username"
-      required="true" matches="[a-z0-9\._-]*" maxlength="16" />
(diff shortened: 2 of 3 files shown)


toKrause (commented):
The same field has the attribute `i18n.error="component.user2.admin.inputhints.username"`.

Currently, `component.user2.admin.inputhints.username` translates to 
`Benutzerkennungen d\u00FCrfen nur maximal 16 Zeichen lang sein und nur aus Kleinbuchstaben und/oder Zahlen bestehen. Umlaute, Sonderzeichen und Leerzeichen sind nicht erlaubt.` and `For user accounts please enter maximum 16 characters only lower case or numbers. Umlauts, special characters and blanks are not allowed!` respecively. But some special charactes (`.`, `_`, `-`) are allowed.

Could you adjust the error message to better reflect the actual requirements as part of this change?

toKrause:
> Could you adjust the error message to better reflect the actual requirements as part of this change?

Never mind, those trnaslations are part of MyCore, not MIR, so a seperate PR is needed anyways.

toKrause (approved):

rsteph-de (changes requested):
Please put english message to proper position

mir-module/src/main/resources/config/mir/messages_en.properties
```diff
@@ -864,3 +864,4 @@ subselect.category.cancel = Cancel
 subselect.category.title  = Choose category
 
 user.profile.id.orcid = Your ORCID iD
+mir.register.username.placeholder = maximum of 16 characters, allowed characters are a-z, 0-9 and ._-
```
toKrause, line 867: New entry should be at the proper location in the file, just as in the German version.

sebhofmann:
maybe try to use @tooltip instead of @placeholder

Possommi:
> Please move english message property to proper position

I did it but I think that is just academic. The line number still don't match since the message properties files are topsy turvy. They should be sorted.

mir-module/src/main/resources/config/mir/messages_en.properties
Possommi: That is just academic. The line number still don't match since the message properties files are topsy turvy. They should be sorted.
I changed the location anyway.

kkrebs (approved):

PR #1071, opened 2024-10-11
MIR-1359 fix choose publication button in related item

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1359).

### DIFF
--- a/mir-module/src/main/resources/META-INF/resources/js/mir/relatedItem-modal.js
+++ b/mir-module/src/main/resources/META-INF/resources/js/mir/relatedItem-modal.js
@@ -1,354 +1,357 @@
-$(document).ready(function() {
-  var GenreXML;
+$(document).ready(function () {
+    var GenreXML;
 
-  $(".mir-related-item-search").each(function() {
-    const id = $(this).find("input[name$='@xlink:href']").val();
-    if (id.length > 0) {
-      const titleInput = $(this).find("input[name*='mods:title']");
-      const urlInput = $(this).find("input[name*='mods:url']");
-      const genreInput = $(this).find("select[name*='mods:genre']");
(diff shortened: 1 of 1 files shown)


eckstueck (approved):

mir-module/src/main/resources/META-INF/resources/js/mir/relatedItem-modal.js
golsch, line 1: Why was the indentation changed? 2-space indentation is actually the standard for JavaScript.

golsch (changes requested):

mir-module/src/main/resources/META-INF/resources/js/mir/relatedItem-modal.js
Antoniafriedrich: You are correct. It seems I hadn't the codestyle settings for mycore javascript imported correctly for some reason. I imported them and autoformated again.

golsch: It is currently still at 4 spaces. According to the documentation, that seems to be okay, but we should possibly question it again. Could you perhaps split the commit and first commit the fix? That way, the changes will be much easier to follow for the review.

Antoniafriedrich: Sure, the diffs are shown a bit confusingly. The only change is in line 5-7 with a new if statement, which is closed in line 24. The rest is just formatting according to mycore-javascriptstyle.xml.

golsch (commented):
The fix is fine, but the new formatting or guidelines for formatting should be reconsidered.

PR #1077, opened 2024-10-25
minor updates to runtime

yagee-de:


sebhofmann (changes requested):
The docker solr version in `mir-webapp/src/main/solr/Dockerfile` should always be the same version as the solr-runner plugin

PR #1078, opened 2024-10-25
MIR-1303 Add handling for non-validatable PDF documents

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1303).

Requires https://github.com/MyCoRe-Org/mycore/pull/2299


sebhofmann:
Still missing the notNull: or exception: extension in the URI-Resolver to prevent unchecked exceptions.

sebhofmann (changes requested):
Still missing the notNull: or exception: extension in the URI-Resolver to prevent unchecked exceptions.

PR #1084, opened 2024-11-12
MIR-1409 Allow to customize badges/icons in search results

Possommi:
[MIR-1409](https://mycore.atlassian.net/browse/MIR-1409)

[MIR-1409]: https://mycore.atlassian.net/browse/MIR-1409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mir-module/src/main/resources/config/mir/mycore.properties
+++ b/mir-module/src/main/resources/config/mir/mycore.properties
@@ -278,6 +278,7 @@ MCR.URIResolver.xslIncludes.mods=%MCR.URIResolver.xslIncludes.mods%,mods2mods-cl
 MCR.URIResolver.xslIncludes.mycoreobjectXML=%MCR.URIResolver.xslIncludes.mycoreobjectXML%,embargofilter.xsl
 MCR.URIResolver.xslIncludes.RDF-mods-journal=%MCR.URIResolver.xslIncludes.RDF-mods-journal%,RDF-mods-journal-mir.xsl
 MCR.URIResolver.xslIncludes.solrResponse=%MCR.URIResolver.xslIncludes.solrResponse%,response-mir.xsl
+MCR.URIResolver.xslIncludes.solrResponseBadges=response-mir-hit-oa.xsl,response-mir-hit-type.xsl,response-mir-hit-licence.xsl,response-mir-hit-date.xsl,response-mir-hit-state.xsl
 MCR.URIResolver.xslIncludes.xeditorTemplates=%MCR.URIResolver.xslIncludes.xeditorTemplates%,xeditor-mir-templates.xsl
 MIR.AVPlayer.Stylesheet=metadata/mir-video.js.xsl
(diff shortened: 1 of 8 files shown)


Possommi:
> Ticket description is about making the icon configurable. If I understand this properly, someone who wishes to change the icon now has to override the new template file and recreate the whole content rendering the badges, coloring, title attribute and all, right?
> 
> I'm not sure that this is the right approach.

It it not about just the icon. In my application there are more states to consider, just on/off wordReadable it not enough. I do not want to fork the whole `mir-response.xsl`.

toKrause:
> It it not about just the icon. In my application there are more states to consider, just on/off wordReadable it not enough. I do not want to fork the whole `mir-response.xsl`.

Please update the JIRA ticket description to better reflect the actual requirements.

That being said, I'm not sure this is a good approach. Having a separate template for just one of the badges. What if someone wants to customize another badge? Should every badge have its own stylesheet? Then, having the code that actually renders a badge copied into many stylesheets doesn't really seem to be a good idea. What about adding additional or removing existing badges? What about badges at different places like the landing pages of publications?

Imho, the badge system would require a complete overhaul. I don't think, that the proposed changes move things in a fruitful direction. 

Changes to the existing system should therefor, for now, be as small and targeted as possible. 

If only the icons need to be configurable, I think, only the icons should be made configurable for now.

Possommi:
> > It it not about just the icon. In my application there are more states to consider, just on/off wordReadable it not enough. I do not want to fork the whole `mir-response.xsl`.
> 
> Please update the JIRA ticket description to better reflect the actual requirements.
> 
> That being said, I'm not sure that having a separate template for just one of the badges is a good approach. What if someone wants to customize another badge? Should every badge have its own stylesheet? Then, having the code that actually renders a badge copied into many stylesheets doesn't really seem to be a good idea. What about adding additional or removing existing badges? What about badges at different places like the landing pages of publications?
> 
> Imho, the badge system desperately requires a complete overhaul. I don't think, that the proposed changes move things in a fruitful direction.
> 
> Changes to the existing system should therefor, for now, be as small and targeted as possible.
> 
> If only the icons need to be configurable, I think, only the icons should be made configurable for now.

Updated the JIRA-Ticket. As I already mentioned, its not just the icon. At the moment the icon is rendered on just one solr field "worldReadableComplete". But in my application I check for other fields as well. Subsequently every part of the hit rendering code should be customizable. But I do not like to do that now as I currently just need to evaluate other solr index data to display a proper icon. I really do not see any problem with providing an extra template/stylesheet rendering that part of the page. 

The problem with MIR is that many stylesheets are huge af and there not much room for customization besides forking the whole stylesheet the same goes for evaluating of access rights but that is another strory.

golsch:
The basket should also be considered, or rather result and basket should use the same stylesheets. Currently the whole thing is inconsistent af.

toKrause:
The badges shown in an object response page should also be harmonized with the badges in a search result page (and the basket listing). Also the use of tooltips, and mouseover effect like color changes and borders should be harmonized. 

Also, the generated HTML should be the same, so that just on set of CSS rulles needs to be maintained.

Also, the selection and order of badges should be configurable.

All in all, there should be a much larger re-engineering of the whole badge mechanism. This is definitely on my list and I have some ideas, but so far I didn't have the time to do it. But, in the meantime, I think it is okay to allow a PR that just tinker around a bit, as long as the changes are somewhat  useful.

In the now proposed solution, I don't like that the invocation of the templates for the various badges is hard coded in `response-mir`xsl`. The original proposal was a bit more flexible here, in that the whole badge rendering was moved to a separate stylesheet. I liked that better.

Possommi:
Implementation in https://github.com/MyCoRe-Org/mir/pull/1150

PR #1094, opened 2024-12-09
MIR-1414 Allow to add extra badges on MIR metadata page

levshyn:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1414).


Possommi:
The fun continues here https://github.com/MyCoRe-Org/mir/pull/1150

PR #1102, opened 2025-01-15
MIR-1419-Can't edit author in publication and related item in same xeditor

sebhofmann:
- also use mods index and relatedItem index in name attribute to identify input elements, to prevent confusion with name index

[Link to jira](https://mycore.atlassian.net/browse/MIR-1419).


kkrebs (approved):
works now!

PR #1108, opened 2025-02-13
MIR-1393 PMD rule MissingSerialVersionUID

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1393).

Also added missing `@Serial` annotation to each `serialVersionUID` and unified the values to `1L`, since serialization isn't used in MyCoRe (in a way hat needs class versioning), only a hand full of classes had actual values and the values were never updated when implementations changed.


yagee-de:
please rebase to 2024.06.x

PR #1116, opened 2025-02-21
MIR-1432 MyCoRe Viewer is missing important stylesheet

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1432).


toKrause:
Not sure if a fix in MyCoRe is really necessary. I don't actually see any reason to use a parameter collector in  `MIRViewerConfigurationStrategy`, since it was only used to look up some values from the configuration and was then discarded; not used with a transformer.

Asserting that `MCRParameterCollector` is only meant to work with a transformer should be okay.

sebhofmann (approved):
This looks good anyway. Properties should be read from MCRConfiguration2 and not from the parameter collector.

Mewel (approved):

PR #1122, opened 2025-03-04
Update to actions/cache@v4

toKrause:


kkrebs:
do we need this update also in 2023.06.x ?

toKrause:
> do we need this update also in 2023.06.x ?

Probably depends on why you're asking, could also reasonably target `2022.06.x` or even `2021.06.x` (since there is at least one [open PR for that branch](https://github.com/MyCoRe-Org/mir/pull/779)). I have no strong opinion here, just that the issue needs to be dealt with urgently.

sebhofmann (approved):

PR #1129, opened 2025-03-13
MIR-1439 Reflect change of attribute names in template mir.languageMenu

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1439)

https://github.com/MyCoRe-Org/mycore/pull/2452


sebhofmann:
This seems like a bug in MyCoRe the field names should stay the same if they are used when serializing, or the serializing should be adopted.

Possommi:
> This seems like a bug in MyCoRe the field names should stay the same if they are used when serializing, or the serializing should be adopted.

@Mewel @sebhofmann so `org.mycore.datamodel.language.MCRLanguageXML` must be adopted?

PR #1156, opened 2025-05-05
MIR-1448 Set defineOwnIdentifier prop in App.vue to true based on DOM attribute

maria-tn:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1448).


sebhofmann (approved):

kkrebs:
can we set "define-own-identifier" per default to true in MIR?

sebhofmann (changes requested):
There is another place in 586, which also needs to be changed to `true`.
```
      <xed:bind xpath="mods:displayForm">
        <input type="hidden"/>
        <div class="nameSearchApp" define-own-identifier="false">
        </div>
      </xed:bind>
```

sebhofmann (approved):

PR #1157, opened 2025-05-06
MIR-1449 Erroneous association of some file extensions with icons in user interface

levshyn:
In the user interface, when viewing a document in the Files section of the derivative, some extension files are incorrectly associated with icons. Example: the .m extension is associated with an icon for audio files.

[Link to jira](https://mycore.atlassian.net/browse/MIR-1449).


sebhofmann (changes requested):
The json file and the resource can be replaced by `MCRConfigHelperServlet`. 

```
  <servlet>
    <servlet-name>MimeTypeIconMappingConfigHelper</servlet-name>
    <servlet-class>org.mycore.frontend.servlets.MCRConfigHelperServlet</servlet-class>
    <init-param>
      <param-name>Properties</param-name>
      <param-value>MIR.Mimetype.IconMapping.*</param-value>
    </init-param>
  </servlet>
  <servlet-mapping>
    <servlet-name>MimeTypeIconMappingConfigHelper</servlet-name>
    <url-pattern>/mime-type-icons.json</url-pattern>
  </servlet-mapping>
```

Property example:
```
MIR.Mimetype.IconMapping.fa-file-code=application/vnd.wolfram.mathematica.package
```

or

Property example:
```
MIR.Mimetype.IconMapping.application/vnd.wolfram.mathematica.package=fa-file-code
```

This does not require a new additional json configuration file and a java resource for fixing a bug.

levshyn:
_This does not require a new additional json configuration file and a java resource for fixing a bug._ -> Yes, that's true. But what if the user needs to change/add a lot of icon mappings and contentType(s)? Will they have to add a lot of **MIR.Mimetype.IconMapping.*** variables? Wouldn't it be easier to just add/change the data in the JSON configuration file?

levshyn:
The idea was to allow the user to simply and easily change the mapping icon -> contentType if needed.

sebhofmann:
> _This does not require a new additional json configuration file and a java resource for fixing a bug._ -> Yes, that's true. But what if the user needs to change/add a lot of icon mappings and contentType(s)? Will they have to add a lot of **MIR.Mimetype.IconMapping.*** variables? Wouldn't it be easier to just add/change the data in the JSON configuration file? The idea was to allow the user to simply and easily change the mapping **icon -> contentType** if needed and not just fix a bug.

You can also inverse your argument. If the User just wants to change a single icon, then he needs to copy the json file, just to change one icon. With my approach, you just need to change one property. 

In MyCoRe we already use properties to configure the application behaviour. It is well a well documented feature and every MyCoRe user knows it. We add additional code to avoid secondary configuration files like persistence.xml. Adding extra code to support a new undocumented configuration file, seems like big step backward.

levshyn:
I agree with your arguments. I'll rework it to read variables from mycore.properties, as you advised.

sebhofmann (changes requested):
Sorry, for the late response. I would prefer another change. Currently there is a mapping from the property to a icon definition e.G. from MIR.Mimetype.IconMapping.Code=application/vnd.wolfram.mathematica.package

But Code is not directly the fontawesome icon, its just a entry in the predefined table `fileIcons`. This allows only to use icons defined in the table and it also increases the code complexity, because its another mapping. I would prefer to not have this mapping and just have a property like:
`MIR.Mimetype.IconMapping.fa-file-code=application/vnd.wolfram.mathematica.package`

This has the following advantages:
- easier to understand
- you can use other icons not present in the js file
- all icons would be in the mycore.properties file

levshyn:
ok, and how should we determine which tooltip to use? At the moment we have const fileIcons in derivate-fileList.js with JSON structure, which uses following definitions as keys: "PDF", "Archive", "Image"... These keys are currently used as tooltips.

sebhofmann:
The labels are wrong now anyway. If i open the site as a german user i get the label "Image". The best way would be to load all keys with a prefix e.G `mir.file.mimetype.label.*` with the translate resource and lookup the right labels. First look for the exact type `mir.mime.label.application.vnd.wolfram.mathematica.package` (replace special characters with .)  and if nothing is found look for generic label `mir.mime.label.application`.

kkrebs:
already fixed as part of #1195

PR #1159, opened 2025-05-08
MIR-1450 Display of badges ignores the version history of metadata 

Possommi:
[MIR-1450](https://mycore.atlassian.net/browse/MIR-1450)

### DIFF
--- a/mir-module/src/main/resources/xsl/badges/mir-badges-entry-point.xsl
+++ b/mir-module/src/main/resources/xsl/badges/mir-badges-entry-point.xsl
@@ -13,4 +13,14 @@
     </span>
   </xsl:template>
 
+  <xsl:template match="/mycoreobject">
+    <xsl:apply-templates select="." mode="mycoreobject-badge"/>
+  </xsl:template>
+
+  <xsl:template match="mycoreobject" mode="mycoreobject-badge">
+    <span>
+      <xsl:apply-imports/>
+    </span>
+  </xsl:template>
+
 </xsl:stylesheet>
--- a/mir-module/src/main/resources/xsl/badges/mir-badges-state.xsl
+++ b/mir-module/src/main/resources/xsl/badges/mir-badges-state.xsl
@@ -18,4 +18,6 @@
       </xsl:call-template>
     </xsl:if>
   </xsl:template>
+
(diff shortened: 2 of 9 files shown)


mir-module/src/main/resources/xsl/modsdetails-external.xsl
fluetze, line 1: Hmm, this calls the URI xslStyle:[something]:notnull:solr:[something]
which means that displaying metadata details page triggers a SOLR query?
Even if we display a historical version? Even if we already currently display an <mcrobject> via /receive?
Does this get the fields back from SOLR we moved there via XSL.Style=solrdocument?

mir-module/src/main/resources/xsl/metadata/mir-abstract.xsl
fluetze, line 1: Does this call solr only for a hit displayed from a result list, or in general?
When metadata is displayed via /receive/ URL, does it use SOLR? 
Why not use xslStyle:badges/mir-badges-entry-point:notnull:mcrobject:', $objectID in that case too?
Why not include mir-badges-entry.point.xsl in this file and apply-templates select="." or "/mcrobject" to get rid of the external call at all?

mir-module/src/main/resources/xsl/badges/mir-badges-state.xsl
fluetze, line 1: Seems to be common in MyCoRe to call org.mycore.common.xml.MCRXMLFunctions:getDisplayName:ClassifID:CategID instead of using the classification resolver?

mir-module/src/main/resources/xsl/badges/mir-badges-entry-point.xsl
fluetze, line 1: Looks ugly, but ok...

mir-module/src/main/resources/xsl/modsdetails-external.xsl
Possommi: this only executed on basket page

mir-module/src/main/resources/xsl/metadata/mir-abstract.xsl
Possommi: The solr resolver is invoked on the metadata page only and not on the result set page (since that page is already in the correct `mode`). Since we have now a code path for displaying badges for history data one could use that for the current object data too.

PR #1194, opened 2025-07-31
MIR-1474 State-badge causes error, when state-field is missing in Solr data

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1474).


toKrause:
It's only really needed because of [MCR-3326](https://github.com/MyCoRe-Org/mycore/pull/2592) in combination with a [now-non-defalt-configuration](https://github.com/MyCoRe-Org/mycore/pull/2592/commits/363ab7291b8a079413d7892c1977cca872b16313).

PR #1227, opened 2025-11-10
MCR-3551 Allow hiding of publicly available derivates

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3551).


mir-module/src/main/java/org/mycore/mir/authorization/MIRStrategy.java
```diff
@@ -162,11 +163,12 @@ private boolean checkDerivatePermission(MCRObjectID derivateId, String permissio
...
         }
 
-        // 1. check if the object has a assigned identifier
+        // 1. check if the object has an assigned identifier and the derivate is exported
         if (MCRAccessManager.PERMISSION_WRITE.equalsIgnoreCase(permission) ||
             MCRAccessManager.PERMISSION_DELETE.equalsIgnoreCase(permission)) {
             final boolean hasRegisteredPI = hasRegisteredPI(objectId);
-            if (hasRegisteredPI && !canEditPI()) {
+            final boolean isExported = MCRXMLFunctions.isDerivateDisplayEnabled(derivateId.toString(), "export");
+            if (hasRegisteredPI && isExported && !canEditPI()) {
```
golsch, line 171: I am not sure about this. What about: `if ((hasRegisteredPI && !canEditPI()) || isExported)`?

mir-module/src/main/resources/META-INF/resources/editor/editor-derivate.xed
```diff
@@ -27,6 +27,9 @@
                              ref="derivate_types"/>
                 <xed:include uri="xslStyle:editor/mir2xeditor:webapp:editor/editor-derivate-includes.xed"
                              ref="mir_access"/>
+                <xed:if></xed:if>
```
golsch, line 30: Typo? But I'm not sure right now how powerful xed is, but can we move the condition `$MIR.Editor.Derivate.EditableIntentServiceFlags != 'false'` here instead of the flags template?

golsch (changes requested):

mir-module/src/main/java/org/mycore/mir/authorization/MIRStrategy.java
toKrause: Wouldn't that mean that any exported derivate (which, by default, would include all derivates), could never be modified?

**Previous logic was:**  
Derivates for objects with registered PI can not be modified (unless user has special permission)

**New logic is intended to be (_same as before, but relaxed for non-exported derivates_):**
Exported derivates for objects with registered PI can not be modified (unless user has special permission)

mir-module/src/main/resources/META-INF/resources/editor/editor-derivate.xed
toKrause: done

mir-module/src/main/java/org/mycore/mir/authorization/MIRStrategy.java
toKrause: so I would like to keep it as is, but I improved the comment

golsch: Sorry, you're absolutely right...

golsch (approved):

PR #1337, opened 2026-04-01
MIR-1568 Fix nameIdentifier bugs in MIR

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1568).

### DIFF
--- a/mir-module/src/main/resources/xsl/mir-mods-utils.xsl
+++ b/mir-module/src/main/resources/xsl/mir-mods-utils.xsl
@@ -66,7 +66,7 @@
                             <xsl:value-of select="@label"/>
                         </dt>
                         <dd>
-                            <a href="{@uri}{translate(@id, ' ', '')}">
+                            <a href="{@uri}{@id}">
                                 <xsl:value-of select="@id"/>
                             </a>
                         </dd>
--- a/mir-module/src/main/resources/xsl/mir-mods-utils.xsl
+++ b/mir-module/src/main/resources/xsl/mir-mods-utils.xsl
@@ -66,7 +66,7 @@
                             <xsl:value-of select="@label"/>
                         </dt>
                         <dd>
-                            <a href="{@uri}{@id}">
+                            <a href="{@uri}{translate(@id, ' ', '')}">
                                 <xsl:value-of select="@id"/>
                             </a>
                         </dd>
--- a/mir-webapp/src/main/vue/name-search/src/App.vue
+++ b/mir-webapp/src/main/vue/name-search/src/App.vue
@@ -29,7 +29,11 @@
               <div class="col-4">
                 <select v-model="model.currentOwnIdentifierType" class="form-control ">
                   <option selected="selected" value="">{{ model.selectLabel }}</option>
-                  <option v-for="identifierType in model.possibleIdentifierTypes.toLowerCase()" :key="identifierType.value" :value="identifierType.value">
+                  <option
+                      v-for="identifierType in model.possibleIdentifierTypes"
+                      :key="identifierType.value"
+                      :value="identifierType.value.toLowerCase()"
+                  >
                     {{ identifierType.label }}
                   </option>
                 </select>
--- a/mir-webapp/src/main/vue/name-search/src/App.vue
+++ b/mir-webapp/src/main/vue/name-search/src/App.vue
@@ -50,7 +50,7 @@
             </div>
             <hr/>
           </template>
-          <div v-for="identifier in model.currentIdentifier.filter(id => isAllowedIdentifierType(id.type))" :key="`${identifier.type}-${identifier.value}`"
+          <div v-for="identifier in model.currentIdentifier" :key="`${identifier.type}-${identifier.value}`"
                v-on:click.prevent="removeIdentifier(identifier)"
                class="identifier">
             <IdentifierDisplay :type="identifier.type" :value="identifier.value"/>
@@ -389,6 +389,9 @@ const addIdentifier = async (identifier: Identifier, event?: MouseEvent) => {
     }
 
   const normType = normalizeType(identifier.type);
+  if (!isAllowedIdentifierType(normType)) {
+    return;
+  }
   if (
       !model.currentIdentifier.some(
           (id: any) => id.type?.toLowerCase() === normType.toLowerCase()
--- a/mir-webapp/src/main/vue/name-search/src/App.vue
+++ b/mir-webapp/src/main/vue/name-search/src/App.vue
@@ -29,7 +29,7 @@
               <div class="col-4">
                 <select v-model="model.currentOwnIdentifierType" class="form-control ">
                   <option selected="selected" value="">{{ model.selectLabel }}</option>
-                  <option v-for="identifierType in model.possibleIdentifierTypes" :key="identifierType.value" :value="identifierType.value">
+                  <option v-for="identifierType in model.possibleIdentifierTypes.toLowerCase()" :key="identifierType.value" :value="identifierType.value">
                     {{ identifierType.label }}
                   </option>
                 </select>
@@ -48,7 +48,7 @@
           <div v-for="identifier in model.currentIdentifier" :key="`${identifier.type}-${identifier.value}`"
                v-on:click.prevent="removeIdentifier(identifier)"
                class="identifier">
-            <identifier-display :type="identifier.type" :value="identifier.value"/>
+            <IdentifierDisplay :type="identifier.type" :value="identifier.value"/>
             <i class="identifier-remover fas fa-minus-circle text-info" v-on:click="addIdentifier(identifier)"></i>
           </div>
 
@@ -94,7 +94,7 @@
                         <div class="col-12">
                             <div v-for="identifier in searchResult.identifier" :key="`${identifier.type}-${identifier.value}`"
                                  class="identifier">
-                                <identifier-display :type="identifier.type" :value="identifier.value"/>
+                                <IdentifierDisplay :type="identifier.type" :value="identifier.value" />
                                 <i :title="model.addLabel"
                                    class="identifier-add fas fa-plus-circle text-info"
                                    v-on:click="addIdentifier(identifier, $event)"></i>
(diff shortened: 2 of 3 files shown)


mir-module/src/main/resources/xsl/mir-mods-utils.xsl
```diff
@@ -66,7 +66,7 @@
                             <xsl:value-of select="@label"/>
                         </dt>
                         <dd>
-                            <a href="{@uri}{@id}">
+                            <a href="{@uri}{translate(@id, ' ', '')}">
```
golsch, line 69: We should discuss this in telco to see if this is a good idea, or if it would be better to store the ids in a machine-readable format in the metadata. There may also be ids of a different type that now behave differently due to the missing spaces.

mir-webapp/src/main/vue/name-search/src/App.vue
```diff
@@ -48,7 +52,7 @@
           <div v-for="identifier in model.currentIdentifier" :key="`${identifier.type}-${identifier.value}`"
                v-on:click.prevent="removeIdentifier(identifier)"
                class="identifier">
-            <identifier-display :type="identifier.type" :value="identifier.value"/>
+            <IdentifierDisplay :type="identifier.type" :value="identifier.value"/>
```
golsch, line 55: This is a stylistic adjustment; generally, `kebap-case` works fine. However, `PascalCase` is recommended to distinguish the template from the standard DOM templates.

mir-webapp/src/main/vue/name-search/src/App.vue
```diff
@@ -260,7 +264,7 @@ const initializeXEditorConnection = async () => {
 
 const convertCategory = (categ: any) => {
     return {
-        value: (categ.ID as string).toLowerCase(),
+        value: (categ.ID as string),
```
golsch, line 267: What is the intention here? I couldn't find any usage where the original case matters, but in all usages, lowercase is relevant.

mir-webapp/src/main/vue/name-search/src/App.vue
```diff
@@ -355,9 +359,9 @@ const addIdentifier = async (identifier: Identifier, event?: MouseEvent) => {
         await animation.finished;
     }
 
-    if (model.currentIdentifier.filter((id: any) => id.type == identifier.type && id.value == identifier.value).length == 0) {
-        model.currentIdentifier.push(identifier);
-    }
+  const normType = (model.possibleIdentifierTypes.find(t => t.value?.toLowerCase() === identifier.type?.toLowerCase())?.value) || identifier.type;
+  if (model.currentIdentifier.filter((id: any) => id.type?.toLowerCase() === normType.toLowerCase() && id.value === identifier.value).length == 0)
+    model.currentIdentifier.push({ ...identifier, type: normType });
```
golsch, line 364: When is the fallback relevant? In general, the behavior changes. But I'm not sure that could even be the case... Why not just leave it as is and filter types without case?

golsch (changes requested):

mir-module/src/main/resources/xsl/mir-mods-utils.xsl
Antoniafriedrich: I understand that this would be better and to format them according to their type when displayed but this is what was generally done in other use cases when working with the identifiers, them being saved in their original format (for example: isni is 0000 0005 0489 765X but for the link we want no spaces).

You can find implementation of similiar logic here:
https://github.com/MyCoRe-Org/mir/blob/dd90c55ba7e91143d35ea20c671d385d39e196f1/mir-webapp/src/main/vue/editor-tools/src/components/display/name-identifier-display.vue#L45-L46

Messing with how we save data could also cause issues to data saved the old way. I think this is important to fix but should not be discussion for a bugfix.

mir-webapp/src/main/vue/name-search/src/App.vue
Antoniafriedrich: When later working with comparing categories this case becomes relevant, because I decided to change the logic here. 

In the original code the logic was:

See what type is in the javascript object. The type is wikidata for example. However we have cases in our classification where this does not match. For example the id for wikidata there is Wikidata, the id for isni is ISNI.

https://github.com/MyCoRe-Org/mycore-classifications/blob/edb125bcb1e7c364437266b4ab1dd6d460b1f475/src/main/resources/mycore-classifications/nameIdentifier.xml#L37

While I think that we should fix the classification, I also think we should not, if available, save the type we found in the javascript, but make sure the id from the classification is saved instead. 

So I changed the logic in App.vue (your other comment) to see if we have in our classification that Id (wikidata from javascript object would match Wikidata from classification case insensitive). If that is the case we save Wikidata instead of wikidata in the type. If not then we save the type as before. 

This is in short the reason why I need the classification ids in their original case for comparison.
The metadata page displays name identifiers very case sensitive by the way, so if I just saved it as type="wikidata" it will not display the name identifier at all.

We still need to update the nameIdentifier.xml because grid is missing completely and multiple ids are not lowercase but I wanted to make sure we stay true to our classification ids first, because this is really messy otherwise.

mir-webapp/src/main/vue/name-search/src/App.vue
Antoniafriedrich: I renamed it so it would be the same to IdentifierDisplay.vue

https://github.com/MyCoRe-Org/mir/blob/dd90c55ba7e91143d35ea20c671d385d39e196f1/mir-webapp/src/main/vue/name-search/src/components/IdentifierDisplay.vue#L19-L21

I saw that other filenames that were used similiarly had kebap-case so I figured the file here was renamed at some point and that its use-cases not being renamed too could cause confusion or problems in the future.

mir-webapp/src/main/vue/name-search/src/App.vue
Antoniafriedrich: Because we now work with the original ids of the classifications. 

Before we had them be transformed to lowercase. So even if our classification file said "Wikidata" we could only work with "wikidata" (I took out .toLowercase() to work with the original ids".

So now comparing id.type and normType (wikidata and Wikidata) would say this is not the same so we just save the nameidentifier as wikidata. We have no classification id for that. It will not get displayed. So we purposefully let the user choose a nameIdentifier in the editor that never shows up in the metadata view.

So I said let us compare it in lowercase (wikidata and Wikidata now is the same) and we save it as type Wikidata which our system knows.

Antoniafriedrich:
**Result of discussion in telko:**
ISNI should be saved without spaces and App.vue will not accept nameidentifiers unknown to our classification. The nameidentifier grid will be added to mycore.classification nameidentifier.xml.

The search will not allow to add identifiers abscent from the classification. The changes made in the xsl will be undone. This is not where we want to change the behavior. 
If we want certain nametypes displayed differently (ISNI) styling choices might be made in the future.

Antoniafriedrich:
Grid will not be added, as Grid nameIdentifier is deprecated. Instead the last commit removes identifiertypes that are not part of the classification from the search component before they are shown to the user and the ISNI is formated in the searcher itself so users "get what they see".

mir-webapp/src/main/vue/name-search/src/App.vue
```diff
@@ -355,9 +364,9 @@ const addIdentifier = async (identifier: Identifier, event?: MouseEvent) => {
         await animation.finished;
     }
 
-    if (model.currentIdentifier.filter((id: any) => id.type == identifier.type && id.value == identifier.value).length == 0) {
-        model.currentIdentifier.push(identifier);
-    }
+  const normType = (model.possibleIdentifierTypes.find(t => t.value?.toLowerCase() === identifier.type?.toLowerCase())?.value);
+  if (model.currentIdentifier.filter((id: any) => id.type?.toLowerCase() === normType.toLowerCase() && id.value === identifier.value).length == 0)
+    model.currentIdentifier.push({ ...identifier, type: normType });
```
golsch: `normType` may can be undefined and the filter will throw an error: use an early return or make the filter more robust.
Also, using `.some()` is both cleaner and more efficient than `filter(...).length == 0`.
Please fix indention and use `{}` for the if block.

mir-webapp/src/main/vue/name-search/src/App.vue
```diff
@@ -92,9 +96,9 @@
                     </div>
                     <div class="row">
                         <div class="col-12">
-                            <div v-for="identifier in searchResult.identifier" :key="`${identifier.type}-${identifier.value}`"
```
golsch, line 95: may fix indention.

mir-webapp/src/main/vue/name-search/src/api/RorSearchProvider.ts
```diff
@@ -57,11 +57,11 @@ export class RorSearchProvider implements SearchProvider {
             for (const extIdObj of item.external_ids) {
                 const idType = extIdObj.type;
                 if (extIdObj.preferred) {
-                    searchResult.identifier.push({ type: idType, value: extIdObj.preferred });
+                    searchResult.identifier.push({ type: idType, value: idType === "isni" ? extIdObj.preferred.replace(/\s+/g, "") : extIdObj.preferred });
```
golsch: may centralize the normalization logic in a helper function.

golsch (changes requested):

sebhofmann:
The GUI currently is inconsistent. Saving an identifier via search result stores it in the original classification case (e.g. Wikidata), while saving it manually via the dropdown stores it lowercase because of the `.toLowerCase()` at App.vue:35. Before this PR everything was lowercase.

 A cleaner solution would be a single helper:
```
 const normalizeType = (raw: string): string => {
     const canonical = model.possibleIdentifierTypes.find(t => t.value.toLowerCase() === raw?.toLowerCase());
     return canonical?.value ?? raw;
 }
```
 - addIdentifier and addOwnIdentifier should both use this helper
 - App.vue:32-36: drop the .toLowerCase() in the option :value so the dropdown emits the same case as the helper.

Antoniafriedrich:
Hope I did not forget anything. The last commit should add the changes asked for. It seemed that indention was not in .vue  from the javascript mycore style guide before so I did not fix that by hand but removed unncessary changes  that have only fomating diffs and no changes.

Since the user is able to add multiple name identifiers of the same type but after clicking the save button only one of each type is saved, I decided to grey out the plus buttons and drop down options if a name identifier of that type is already existing.
The user's view of what they can do should match what they will get.

golsch (approved):
Looks good. In the future, we should reformat all TS/Vue files consistently.

PR #1339, opened 2026-04-02
MIR-1566 Redesign Affiliation Field in Submission Forms

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1566).


TiloNN:
Uh .. please do not use @extend in Sass. To make it easy: Never.
Seems a very cool feature but makes more troubles than helping us.
Short:
- blows up CSS Code fast
- very hard to maintain 

Just use the bs utility classes in HTML Code instead of inheriting them in sass.

TiloNN:
Nice!

Here some Ideas for further improvement:
- use the same background color as disabled elements for input and badge container
`background-color: var(--bs-secondary-bg);` or $body-secondary-bg
- make the complete badge clickable 
- change the cursor on hovering the badge
`cursor: pointer;`
- remove the shadow from the input when it gets focus
`inputselector:focus {box-shadow: none;}`

optional:
- prevent that the input has a pointer by removing focus when it gets one
`onfocus="blur();"`
But I'm not sure if this is clean webdesign. Seems a little bit dirty.

I can try to add those improvements by my own.
But feel free, if you don't want to wait.

TiloNN (approved):
No ideas for more improvements.
Thanks and well done.

PR #1352, opened 2026-04-28
MIR-1573 Add Solr fields for funding information

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1573).

### DIFF
--- a/mir-module/src/main/resources/config/mir/solr/main/solr-schema.json
+++ b/mir-module/src/main/resources/config/mir/solr/main/solr-schema.json
@@ -249,5 +249,50 @@
         "stored": true,
         "multiValued": false
     }
+  },
+  {
+    "add-field": {
+      "name": "funding_funder_name",
+      "type": "string",
+      "indexed": true,
+      "stored": true,
+      "multiValued": true
+    }
+  },
+  {
+    "add-field": {
+      "name": "funding_funder_identifier",
+      "type": "string",
+      "indexed": true,
+      "stored": true,
+      "multiValued": true
+    }
+  },
+  {
+    "add-field": {
+      "name": "funding_award_title",
+      "type": "string",
+      "indexed": true,
+      "stored": true,
+      "multiValued": true
+    }
+  },
+  {
+    "add-field": {
+      "name": "funding_award_number",
+      "type": "string",
+      "indexed": true,
+      "stored": true,
+      "multiValued": true
+    }
+  },
+  {
+    "add-field": {
+      "name": "funding_award_uri",
+      "type": "string",
+      "indexed": true,
+      "stored": true,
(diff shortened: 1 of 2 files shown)


golsch:
The fields are already being indexed almost _by magic_ via `solr/indexing/mods-solr.xsl` from `mycore-mods`  or am I mistaken?

for example:
```
<field name="awardTitle">Towards a non-invasive molecular test for bladder cancer</field>
<field name="awardNumber">5R01CA116161-09</field>
<field name="funderName">National Institutes of Health</field>
<field name="funderIdentifier">https://doi.org/10.13039/100000002</field>
<field name="funderIdentifier.funderIdentifierType.Crossref_Funder_ID">https://doi.org/10.13039/100000002</field>
<field name="funderIdentifier.funderIdentifierType">Crossref Funder ID</field>
```

toKrause:
Apparently so. I did not know about that feature. I've only ever worked with fields that exist explicitly. Seems to be little/not known by many experts, since nobody mentioned it, when I talked about the need for these new search fields in lasts weeks web call.

---

One question though. **How can the magic fields be used to search for something?**

After starting up a new MIR instance (current `main` branch, via Docker) and creating a published object with funding information and confirming that the corresponding magic fields exist (`resolve uri xslTransform:mycoreobject-solrdocument:mcrobject:mir_mods_00000001`), searching for objects with award numbers (via the expert search) with query `awardNumbe:*` (`http://localhost:8291/mir/servlets/solr/select?core=main&q=awardNumber%3A*&wt=xml`) doesn't yield any results.

Same with, for example, `title:*`. Searching for `mods.title:*` yields results, so the general query syntax seems to be okay (for explicitly existing fields).

toKrause:
This is probably because [all magic field are ignored](https://github.com/MyCoRe-Org/mycore/blob/d360a93491c60eab9fcbaa9b27d37be7909147b8/mycore-solr/src/main/resources/components/solr/config/solr/main/solr-schema.json#L746) (neither indexed nor stored)?

rsteph-de:
Some fields are automatically generated into the SolrInputDocument through XSLT. 
Field names are buidl from element names and attribute values.
They still need to be added to the Solr schema, to be stored in Solr and used in the application. Otherwise Solr will simply ignore them.

toKrause:
If the fields have to be added manually anyways, I think I'd prefer the hand-crafted versions as suggested in this PR, because of:

- nicer names ("namespaced", agreed upon snake_case style)
- better aligned identifier handling

golsch:
I agree with you. But in the future, we should prevent the other fields from magically appearing next to them, even if they end up being dropped because of the schema.

golsch (approved):

PR #1353, opened 2026-04-29
MIR-1574 adds maintenance message feature

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1574).

PR #1357, opened 2026-05-05
MIR-1574 adds a maintenance message feature to main

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1574).

New pr after feedback from 
https://github.com/MyCoRe-Org/mir/pull/1353

### DIFF
--- a/mir-module/src/main/resources/config/mir/mycore.properties
+++ b/mir-module/src/main/resources/config/mir/mycore.properties
@@ -806,5 +806,10 @@ MCR.ContentTransformer.svg-download.Stylesheet = xslt/generate-svg-by-type.xsl
 
 MCR.Startup.Class = %MCR.Startup.Class%,org.mycore.mir.migration.MIRMigrateStaticHistoryContent
 
+##############################################################################
+# Maintenance configuration
+##############################################################################
+MIR.Maintenance.File=%MCR.datadir%/maintenance.xml
+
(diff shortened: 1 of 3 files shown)


sebhofmann (changes requested):
I would prefer if we do not set a default for MIR.Maintenance.File. So we can check for length >0 in the Stylesheet and save one call to the uri resolver if the Property is not set.

mir-module/src/main/resources/config/mir/mycore.properties
```diff
@@ -806,5 +806,10 @@ MCR.ContentTransformer.svg-download.Stylesheet = xslt/generate-svg-by-type.xsl
 
 MCR.Startup.Class = %MCR.Startup.Class%,org.mycore.mir.migration.MIRMigrateStaticHistoryContent
 
+##############################################################################
+# Maintenance configuration
+##############################################################################
+MIR.Maintenance.File=%MCR.datadir%/maintenance.xml
```
toKrause: Why `%MCR.datadir%` rather than `%MCR.basedir%`.

A `maintenance.xml` file seems to me to be more akin to a configuration file than a data file.

TiloNN:
Feedback added.
Please check again.

kkrebs (approved):

toKrause (approved):

sebhofmann (approved):

PR #1368, opened 2026-05-14
MIR-1581 remove obsolete not working resolvePPN javascript function

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1581).


sebhofmann (changes requested):
Should remove `trimPPNUrl` too. Since its private to this file and not used anymore.

sebhofmann (approved):

PR #1398, opened 2026-06-17
MIR-1593 Some stylesheets are mising after xslt migration

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1593).


kkrebs (approved):
successfully tested
🚢

PR #1399, opened 2026-06-17
MIR-1595 de-duplicate genre dc:type and receive-URL dc:identifier in mods2dc

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1595).

## Summary

The DC export (`mods2dc`) produced duplicate values, e.g. three identical `dc:type` `article` plus two identical receive-URL `dc:identifier`. This PR de-duplicates both in `includes/mods2dc.xsl`:

- **genre -> dc:type**: collected per distinct value, so the intern genre and the mapped marcgt genre no longer yield duplicate `dc:type`. The shared `mods:genre` template is kept on purpose — the OAI export (`mods2oai_dc.xsl`) includes this stylesheet and applies `mods:genre`, so it keeps working.
- **dc:identifier**: the enhancer-added receive URL identifier (`mods:identifier` with `type='uri'` pointing to `.../receive/ID`) is filtered out, since the same receive URL is already emitted explicitly.

## Relation

Builds on MyCoRe **MCR-3728** (mycore PR #2990), which removes the redundant marcgt genre in `mods-enhancer.xsl` and brings the duplicate `dc:type` from 3 down to 2. This PR takes it to 1 and additionally fixes the duplicate identifier.

## Testing

Verified locally with Saxon (HE 12.9):
- DC export now yields a single `dc:type` `article`, `Text`, `doc-type:Article`, and a single receive `dc:identifier`; real identifiers (isbn/issn/external uri/hdl) are preserved.
- OAI export (single intern genre) still produces its `dc:type`.
- Two genres with distinct types are both kept (only identical values collapse).

### DIFF
--- a/mir-module/src/main/resources/xslt/includes/mods2dc.xsl
+++ b/mir-module/src/main/resources/xslt/includes/mods2dc.xsl
@@ -9,8 +9,10 @@
   xmlns:srw_dc="info:srw/schema/1/dc-schema"
   xmlns:xlink="http://www.w3.org/1999/xlink"
   xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+  xmlns:mcrmir="http://www.mycore.org/mir/functions"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


kkrebs (approved):
looks good now!

PR #1428, opened 2026-06-24
MIR-1604 Change password page is not marked as active page in menu

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1604).


sebhofmann (changes requested):
Das geht so nicht. Wenn man per Shibboleth angemeldet ist, dann wird man auf die Passwort-Ändern-Seite des IDP weitergeleitet und nicht zu dem XED Formular.

kkrebs:
Thanks for the tip!
In that case, it can’t be fixed. The logic compares the URL in the navigation.xml with the browser URL and then decides which item gets the active marker. So I’ll move the ticket out of the sprint, and we’ll need to come up with another solution. But I don’t think it’s super urgent.

PR #1432, opened 2026-06-25
MIR-1622 Restore embargo filter behavior

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1622).

### DIFF
--- a/mir-module/src/main/resources/xslt/embargofilter.xsl
+++ b/mir-module/src/main/resources/xslt/embargofilter.xsl
@@ -19,7 +19,7 @@
       <xsl:for-each select="derobject">
         <xsl:variable name="derivate-id" select="@xlink:href" />
         <xsl:variable name="current-user-can-read-derivate" as="xs:boolean" select="
-          mcracl:check-permission($derivate-id, 'read') or mcracl:check-permission($derivate-id, 'preview')
+          mcracl:check-permission($derivate-id, 'read') or mcracl:check-permission($derivate-id, 'view')
         " />
         <xsl:if test="$current-user-can-read-derivate or not($current-user-is-guest)">
           <xsl:choose>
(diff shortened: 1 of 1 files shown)


sebhofmann:
The embargo should only be set if the date is in the future. There is a job which changes the embargo to expired embargo.

toKrause:
As far as i can tell, this PR restores the pre-[#1424](https://github.com/MyCoRe-Org/mir/pull/1424)-behavior as closely as possible, which is the goal of this PR. The implementation was unintentionally changed in that PR.

Changing the implementation of the embargo filter seems to be a good idea, but should be a separate PR, not part of [MIR-1622](https://mycore.atlassian.net/browse/MIR-1622). What that implementation should be, should probably be discussed in a future web conference.

[MIR-1622]: https://mycore.atlassian.net/browse/MIR-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

toKrause:
Diff of the original version with the version proposed in this PR (with reformatting, variable-renaming, storing selections as variables, ... removed for better comparison):

Screenshot from 2026-06-25 10-19-33

mir-module/src/main/resources/xslt/embargofilter.xsl
```diff
@@ -10,41 +9,48 @@
...
-          </xsl:when>
-          <xsl:otherwise>
-            <xsl:copy>
-              <xsl:apply-templates select="@* | node()" />
-            </xsl:copy>
-          </xsl:otherwise>
-        </xsl:choose>
+        <xsl:variable name="derivate-id" select="@xlink:href" />
+        <xsl:variable name="current-user-can-read-derivate" as="xs:boolean" select="
+          mcracl:check-permission($derivate-id, 'read') or mcracl:check-permission($derivate-id, 'preview')
```
toKrause: should be `mcracl:check-permission($derivate-id, 'view')`, not `mcracl:check-permission($derivate-id, 'preview')`

toKrause (changes requested):

toKrause (approved):

PR #1436, opened 2026-06-25
MIR-1628 Make sword Saxon compatible

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1628).

relies on https://github.com/MyCoRe-Org/mycore/pull/3009


sebhofmann (commented):
## Local review — MIR-1628 Make sword Saxon compatible

Reviewed the XSLT 1.0 (Xalan) → 3.0 (Saxon) migration of `dissemin-mods2mycore-mods.xsl` and the `mycore.properties` path update. The migration is mostly clean, but there is one behavioral regression.

### 🔴 Regression — `languageTerm` rfc3066 → rfc5646 conversion becomes dead code

The old identity template (`match="@*|node()"` doing a shallow copy + recurse) was replaced by:

```xml
<xsl:mode on-no-match="deep-copy" />
```

`deep-copy` is **not** equivalent to the old identity template. The built-in `deep-copy` rule copies an unmatched node together with its whole subtree and **does not apply templates to its children**, whereas the old identity template recursed into the children.

Consequence: `mods:language` has no explicit template, so it is handled by `deep-copy`. Its child `mods:languageTerm[@type='code' and @authority='rfc3066']` is therefore copied verbatim and the conversion template never fires. The `authority` stays `rfc3066` instead of being rewritten to `rfc5646`. The same happens for `mods:language` nested inside `mods:relatedItem`. The template that does the rfc3066 → rfc5646 rewrite effectively turns into dead code.

**Fix:** use `shallow-copy` instead, which copies the node *and* applies templates to its children — the exact Saxon-compatible equivalent of the removed identity template:

```xml
<xsl:mode on-no-match="shallow-copy" />
```

### 🟡 Minor

- Stray trailing space inside the `licenseID` select expression: `)[1] "`. Harmless, but worth removing.
- The `licenseID` refactor to `( … , … )[1] as xs:string?` is correct and a nice simplification. 👍
- The `string-length(x) > 0` → `x` test changes for `romeoId` / `dissemInID` are equivalent in practice (an empty element has no `text()` node).

### Verdict

Solid migration apart from the `deep-copy` regression, which silently breaks runtime behavior. Recommend switching `on-no-match` to `shallow-copy` before merging.

*(Reviewed locally against the merge-base with `2026.06.x`.)*

sebhofmann (approved):

PR #1444, opened 2026-06-30
Issues/mir 1604 change password entry not active

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-MIR-1604).

Its a hack, but works for me now.
Ideas for better solutions?

### DIFF
--- a/mir-module/src/main/resources/xslt/layout/mir-navigation.xsl
+++ b/mir-module/src/main/resources/xslt/layout/mir-navigation.xsl
@@ -39,7 +39,7 @@
   </xsl:template>
 
   <xsl:template match="/navigation//item[@href]">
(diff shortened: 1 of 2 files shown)


mir-module/src/main/resources/xslt/layout/mir-navigation.xsl
```diff
@@ -66,6 +66,9 @@
           </xsl:if>
         </xsl:variable>
         <li class="{$listItemClass}">
+          <span><xsl:value-of select="$browserAddress"/></span>
```
sebhofmann: remove the "debug" code

TiloNN: Oh gosh, yes.
Thx. Done.

sebhofmann (approved):

PR #1445, opened 2026-06-30
MIR-1633 Fix basket on object page, search hit and basket display

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1633).

The XSLT/Saxon migration broke three parts of the object basket:

- **Object page** (`modsdetails-external.xsl`): the membership check used a `callJava:` resolver whose result is always a non-empty node-set, so the `xsl:when` test was always true and the page always showed the "remove from basket" button. Now uses the `basket:` URI resolver and tests for the entry's `@uri`.
- **Search result hit** (`response-mir.xsl`): the check used basket type `object` (singular) while the servlet uses `objects` (plural). These are separate baskets, so it never matched. Fixed the type.
- **Basket display** (`modsdetails-external.xsl`): the `basketContent` template only matched the document root (`/mycoreobject`), but `action=show` embeds the object as a child of `<entry>`, so the entry was dumped as raw text. Now matches `mycoreobject` in any position. (The object-page fix resolves the entry content into the session via the `basket:` resolver, which is what triggers the embedded path.)

## Tests

Adds `MIRBasketITCase` (Selenium IT) covering:
- add/remove toggle on the object page,
- basket display rendering with the formatted hit layout (not raw text),
- search hit reflecting basket membership,
- entry reordering (up/down).

Full `mvn clean install` incl. integration tests passes locally; `MIRBasketITCase` runs 2 tests, 0 failures.

### DIFF
--- a/mir-it/src/test/java/org/mycore/mir/it/controller/MIRBasketController.java
+++ b/mir-it/src/test/java/org/mycore/mir/it/controller/MIRBasketController.java
@@ -1,0 +1,133 @@
+package org.mycore.mir.it.controller;
+
+import java.util.List;
+
+import org.mycore.common.selenium.drivers.MCRWebdriverWrapper;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+/**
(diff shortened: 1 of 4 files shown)


kkrebs (approved):
successfully tested with local MIR

PR #1459, opened 2026-07-10
MIR-1643 removed or replaced outdated form-inline class

TiloNN:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1643).


kkrebs (approved):
looks good for me, and xed template with id "classification" is obsolete and no longer used in MIR. That's why I think we can delete it.

PR #1461, opened 2026-07-14
MIR-1631 Allow better styling of thumbnails with non-DIN-A-aspect-ratio

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1631).


TiloNN:
Great, you replaced the background images with image elements.
I appreciate that, and it works for me, but I have some questions to better understand your approach.
Perhaps it would be better to comment directly in the code, but I’ll start here.

1. Will we never have background images in the .hit_icon again? If so, there’s no reason to keep the background properties.
2. Why did you replace the height with an aspect-ratio command? To achieve a true DIN A4 format? And is this why you also needed to replace the placeholder images?
3. You decided to use flexbox to center the image and fill the space with width and height properties. It works. Another approach I would have tried is using object-fit and object-position.
4. It seems strange to me that we now use the disabled image when not(mcracl:is-current-user-guest-user()) instead of the default as before. But perhaps I simply don’t understand the code. I even do not know, why we need a disabled and a default image with just a gray color and no content.

All points are non-blockers, aren't they? They are just annotations. 
Perhaps you could extend the good documentation in the ticket.
Uh .. perhaps you just answer there. Will copy the text.

Will short check the changes with Kathleen today and will approve it then.

toKrause:
> * Will we never have background images in the .hit_icon again? If so, there’s no reason to keep the background properties.
> 
> * Why did you replace the height with an aspect-ratio command? To achieve a true DIN A4 format? And is this why you also needed to replace the placeholder images?
> 
> * You decided to use flexbox to center the image and fill the space with width and height properties. It works. Another approach I would have tried is using object-fit and object-position.
> 
> * It seems strange to me that we now use the disabled image when not(mcracl:is-current-user-guest-user()) instead of the default as before. But perhaps I simply don’t understand the code. I even do not know, why we need a disabled and a default image with just a gray color and no content.

- I will remove the now unnecessary `background-*`-properties (except `background-color`).
- The combination of fixed width and fixed aspect-ratio works well to preserve the fixed DIN-A-aspect-ratio in MIR and also allows to opt-out of that constraint easily by simply removing the fixed aspect-ratio (with `aspect-ratio: unset`)
- I'm not a CSS expert, so I don't know about any pros and cons of the two approaches. The one I used appears to be working fine, which is enough for me :)
- `icon_common_disabled.png'` is now used as the interim image until the actual image (referenced in the `data-iiif-jwt`-attribute) is loaded with JS (the `../default.jpg` is a fixed part of the IIIF-URL, but this is indeed the object-specific thumbnail). Otherwise no image would be show for that split-second, if the aspect-ratio has been removed (see point 2), or in cases where the JS fails to load the actual image.

kkrebs (approved):

TiloNN (approved):
Thx!
No more questions :)

PR #1495, opened 2026-08-24
MIR-1655 expand searchproviders with dante connection

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MIR-1655).

### DIFF
--- a/mir-webapp/src/main/vue/editor-tools/README.md
+++ b/mir-webapp/src/main/vue/editor-tools/README.md
@@ -27,3 +27,33 @@ yarn lint
 
 ### Customize configuration
 See [Configuration Reference](https://cli.vuejs.org/config/).
+
+## Provider Setup
+
+As standard to search for topics, the LobidProvider is set. To extend or change this behaviour the user can register 
+other providers using one of the editor templates. To have multiple vocabularies inside one Dante provider, 
+you can seperate them with "|"
+
+```
+          <div class="editorToolsApp"
+               data-searchable="*"
+               data-editor="*"
+               data-search-filter-default="*"
+               data-admin="true"
+               data-required="false"
+               
+               data-provider-lobid-label="Lobid"
+               data-provider-lobid-enabled="true"
+               data-provider-lobid-authority="gnd"
+               data-provider-lobid-baseurl="https://lobid.org/gnd/search"
+
+               data-provider-dante-enabled="true"
+               data-provider-dante-authority="dante"
+               data-provider-dante-label="Dante"
+               data-provider-dante-vocabulary="foo|bar"
+               data-provider-dante-baseurl="https://api.dante.gbv.de/search"
+          
+          >
+          </div>
+```
+Other providers may be added from your project using the SearchProviderRegistry.
--- a/mir-webapp/src/main/vue/editor-tools/README.md
+++ b/mir-webapp/src/main/vue/editor-tools/README.md
@@ -32,7 +32,7 @@ See [Configuration Reference](https://cli.vuejs.org/config/).
 
 As standard to search for topics, the LobidProvider is set. To extend or change this behaviour the user can register 
 other providers using one of the editor templates. To have multiple vocabularies inside one Dante provider, 
-you can seperate them with "|"
+you can seperate them with "|". Alternatively multiple providers of the same type are supported. 
 
 ```
           <div class="editorToolsApp"
@@ -44,16 +44,18 @@ you can seperate them with "|"
                
                data-provider-lobid-label="Lobid"
                data-provider-lobid-enabled="true"
-               data-provider-lobid-authority="gnd"
-               data-provider-lobid-baseurl="https://lobid.org/gnd/search"
+               data-provider-lobid-base-url="https://lobid.org/gnd/search"
 
                data-provider-dante-enabled="true"
-               data-provider-dante-authority="dante"
                data-provider-dante-label="Dante"
+               data-provider-dante-type="dante"
                data-provider-dante-vocabulary="foo|bar"
-               data-provider-dante-baseurl="https://api.dante.gbv.de/search"
-          
+               data-provider-dante-base-url="https://api.dante.gbv.de/search"
+               data-provider-dante-keys="inScheme,notation,altLabel"
+
           >
           </div>
 ```
 Other providers may be added from your project using the SearchProviderRegistry.
+When using the DanteSearchProvider you can add additional properties from the result that should be displayed.
+To display them they need a message language property.
--- a/mir-webapp/src/main/vue/editor-tools/src/api/XEditorConnector.ts
+++ b/mir-webapp/src/main/vue/editor-tools/src/api/XEditorConnector.ts
@@ -34,6 +34,19 @@ export const retrieveSubject = (root: HTMLElement): Subject => {
 
 };
 
+export interface ProviderConfigDetails {
+    authorityName?: string;
+    baseUrl: string;
+    vocabulary?: string;
+    [key: string]: string | undefined;
+}
+
+export interface ProviderConfig {
+    id: string;
+    type: string;
+    config: ProviderConfigDetails;
+}
+
 export interface EditorSettings {
     /**
      * List of types which facet should be enabled in the search.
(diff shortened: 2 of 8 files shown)


sebhofmann (commented):
Reviewed `git diff main...HEAD` (8 files). `tsc --noEmit` is clean, the 15 existing unit tests pass (none of them touch the new code), and the lib build completes.

The overall approach is right. Two things should be resolved before merging: the first one turns the integration tests red, the second one misses the core requirement of the ticket.

## Blocking

### 1. Tab title breaks the integration tests

`mir-webapp/src/main/vue/editor-tools/src/components/editor-wrapper.vue:279,289`

```ts
title: providerConfig.id   // "lobid"
```

This used to be the literal `title: "Lobid"`. The integration tests select the tab by its text:

* `mir-it/src/test/java/org/mycore/mir/it/controller/MIRModsEditorController.java:382` uses `a[contains(text(),'Lobid')]`
* `mir-it/src/test/java/org/mycore/mir/it/controller/MIRModsEditorController.java:417` does the same

XPath `contains()` is case sensitive, so neither location finds the tab any more. There is also a UI regression: the dialog now shows the lowercase technical ID instead of a display name.

Suggestion: take the title from the config (for example `data-provider-lobid-label`) or from an i18n key such as `mir.editor.subject.provider.<id>.title`, falling back to the ID.

### 2. The global registry prevents exactly what the ticket asks for

`mir-webapp/src/main/vue/editor-tools/src/api/search/SearchProviderRegistry.ts:1`, `editor-wrapper.vue:220-226`

`SearchProviders` is a module global `Record<string, any>` keyed by provider ID, and registration is skipped when the ID is already taken:

```ts
if (!SearchProviders[providerConfig.id]) { registerProvider(providerConfig); }
```

`editor-includes.xed` renders many `editorToolsApp` divs per form. Two templates configured with `data-provider-dante-vocabulary="css_color_names"` and `"rvk"` share the ID `dante`, so the first div that mounts wins and the second one silently searches the wrong vocabulary. The ticket explicitly asks for "the possibility to provide each template with a configuration".

Suggestion: hold the instance per editor (registry hands out classes, instances live in the component `model`), or include the config in the cache key.

## Bugs

### 3. Crash on an empty result list

`search-result-list.vue:179-182`

```ts
if (!isGroupPresent) { model.currentGroupId = newValue[0].groupId; }
```

There is no guard for an empty array. This became reachable through the new catch branch in `editor-wrapper.vue:295`, which sets `model.searchResultGroup = []`, so `newValue[0]` is `undefined` and the watcher throws a TypeError. An empty `providers` array leads to the same place.

### 4. `inject("editorSettings")` is dead code

`editor-wrapper.vue:217`

There is no `provide()` anywhere in the project, so the call always returns `undefined` and Vue logs `[Vue warn] injection "editorSettings" not found` in dev builds. Either pass a default or drop the line.

### 5. A failed import poisons the registry permanently

`SearchProviderRegistry.ts:5-38`

`instantiateProvider` has no try/catch. A rejected promise stays in `pendingPromises`, so every later `ensureProvidersLoaded()` call rejects as well. On top of that, the `await ensureProvidersLoaded()` in `onMounted` (`editor-wrapper.vue:225`) is unguarded, so the rest of the mount is skipped: the `searchOptions` defaults are never applied and the submit guard (the `confirm` when leaving the page with a non empty search field) is never registered.

Suggestion: catch errors per provider.

### 6. The spinner gets stuck

`editor-wrapper.vue:263` and `editor-wrapper.vue:269`

`searchSubmitted` sets `model.searching = true`, then `search()` bails out at `if (!model.settings?.providers) return;` before the `finally` block runs, so the loading indicator stays visible forever. The early return should reset `searching`.

### 7. `authorityName` can silently be undefined

`SkosProviderConfig.authorityName` is a required `string`, `ProviderConfigDetails.authorityName` is optional, and `registerProvider(item: {config: any})` removes the type check entirely. If an overlay sets `data-provider-x-enabled="true"` without `-authority`, every hit is written to MODS without an `authority` attribute, because `SubjectXMLOutputter.ts:17` skips it silently. That is a data quality problem which only surfaces months later.

Suggestion: validate the required field, log a `console.error` and discard the provider.

## Design

### 8. The base class knows about GND

`SearchProvider.ts:37-343`

`handleMember`, `handleCorporateBody`, `addGNDLink` and `settingsToQuery` with its `type:SubjectHeading OR type:Work` query are Lobid and GND specifics, but they now live in the abstract base class. `DanteSearchProvider` inherits roughly 250 lines it never uses, and `SkosProviderConfig` is named after SKOS even though Lobid does not speak SKOS.

Suggestion: keep `SearchProvider` minimal (config, `search()`, `generateID()`) and add an intermediate `GndSearchProvider` holding the GND mapping, which `LobidSearchProvider` extends. Visibility is also mixed: `handleMember` and `handleTitle` are public while `handleTopic` and `handlePlace` are protected.

### 9. The attribute name parsing is too narrow

`XEditorConnector.ts:190`

```ts
const match = key.match(/^provider([A-Z][a-z0-9]*)(.*)$/);
```

The provider ID has to be a single word. `data-provider-my-dante-baseurl` yields the ID `my` and the property `danteBaseurl` without any warning. The manual special cases for `baseurl` and `authorityname` point in the same direction.

Suggestion: enforce a separator, for example `data-provider="dante"` plus fixed property names, or cut the ID at the first known property suffix.

In the same block, the spread sits after the explicit keys, so `enabled` and `type` end up inside `config`. It works, but it reads as if the explicit values were meant to win.

### 10. Hardcoded default URL in two places

`https://lobid.org/gnd/search` appears both in the fallback (`XEditorConnector.ts:212`) and in `LobidSearchProvider.ts:15`. That goes against the "no hardcoded defaults" convention, and one of the two will be missed on the next change.

### 11. Undocumented global API

`SearchProviderRegistry.ts:41-53`

After the pending entries are processed, `window.MyCoRePendingProviders` is replaced by an object that only has `push`, so it is no longer an array. Any external code using `forEach` or `length` breaks. `window.MyCoReSearchProviderRegistry` is the extension point for overlays, but it is neither documented nor typed. The `MyCoRe*` prefix is also unusual inside a MIR module.

### 12. Code splitting buys nothing here but adds deployment assumptions

The dynamic `import()` calls produce two extra chunks in the lib build (`editor-tools.umd.min.254.js` and `editor-tools.umd.min.582.js`, about 6 KB each) next to the 1.8 MB main bundle. The public path is derived from `document.currentScript.src` and the Maven build places the chunks in the same directory, so it does work. However, any overlay that copies only `editor-tools.umd.min.js` gets a 404 on the first search, and the saving is not measurable against 1.8 MB. Static imports would be more honest here.

## Dante specifics

`DanteSearchProvider.ts`

* The labels `"URI"`, `"English Label"`, `"German Label"` and `"Notation"` are hardcoded English. All Lobid labels go through `i18n(...)` and have keys in `messages_de.properties` and `messages_en.properties`. Those keys are missing.
* No size limit. Lobid sends `size=30`, while Dante gets `query=*<term>*` with no bound, and even a plain `*` over the whole vocabulary for an empty term.
* `this.baseUrl` has no fallback. With an empty config the request goes out relative to the current page.
* Only `settings.searchTopic` is evaluated, everything else returns empty. That is not extensible for vocabularies containing persons or places, so it deserves at least a comment.
* `*${searchTerm}*` doubles wildcards a user already typed.

## Minor

* **Indentation.** `editor-wrapper.vue:214-229` was switched to 2 spaces while line 230 onwards still uses 8 spaces inside the same block. `search-result-list.vue:18` sits at 10 spaces instead of 16. The conventions ask for 4 spaces in JS.
* **Unrelated XSL change.** `mir-badges-state.xsl` is a behaviour change: a present but empty `state` field no longer renders a badge. That looks correct, but it belongs in its own commit with its own ticket.
* **PR body.** `[Link to jira](https://mycore.atlassian.net/browse/MIR-)` is a dead link, the ticket number is missing, and `PULL_REQUEST_TEMPLATE.md` is not filled in.
* **No tests.** Nothing covers the `retrieveSettings` attribute parsing (that regex is asking for tests), the registry, or the Dante mapping.
* **No documentation.** There is not a single `data-provider-*` example in `editor-includes.xed` or the README. Keeping Lobid as the default is correct per the ticket, but without an example in the repository nobody will find the configuration.

## Nice to see

* The tab click in `search-result-list.vue:18` was a bare `@click.prevent` with no handler, so the tabs were never switchable. That is a real fix which only becomes visible once there is more than one provider.
* `for...in` over arrays replaced by `for...of`, `substr` replaced by `substring`, missing newlines at end of file added.
* The `response.ok` check and the try/catch around the network calls. Previously a Lobid outage could break the search dialog outright.

